PR #7, opened 2017-03-12
Timbl davos

timbl:
Incremental improvements in the meetulator and contacts.  Also new values of termType string with initial caps Collection and Graph

### DIFF
--- a/contact/Makefile
+++ b/contact/Makefile
@@ -1,0 +1,12 @@
+# Wrap TTL files into JS files for bundling with library
+
+,all : individualForm.js vcard.js
+
+individualForm.js : individualForm.ttl
+				(echo 'module.exports = `' ; cat individualForm.ttl; echo '`') >  individualForm.js
+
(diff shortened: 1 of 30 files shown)


contact/Makefile
```diff
@@ -0,0 +1,12 @@
+# Wrap TTL files into JS files for bundling with library
+
+,all : individualForm.js vcard.js
+
+individualForm.js : individualForm.ttl
+				(echo 'module.exports = `' ; cat individualForm.ttl; echo '`') >  individualForm.js
```
dan-f, line 6: So this is one way of exporting arbitrary content from within JS modules. Another way would be to use a tool like `webpack` and to design a custom [loader](https://webpack.js.org/concepts/loaders/) for turtle. Perhaps one even exists.

dan-f (commented):
There's certainly a lot here... :) Based on our earlier discussion, it seems like it's doing what you want :+1:

PR #16, opened 2017-09-13
Fix missing encodeURIComponent in filename to URI translation

timbl:
Fix missing encodeURIComponent in filename to URI translation

### DIFF
--- a/folderPane.js
+++ b/folderPane.js
@@ -8,6 +8,7 @@ var UI = require('solid-ui')
 // var Solid = require('solid-client')
 
 var ns = UI.ns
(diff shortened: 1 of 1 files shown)


folderPane.js
```diff
@@ -8,6 +8,7 @@ var UI = require('solid-ui')
 // var Solid = require('solid-client')
 
 var ns = UI.ns
+console.log('@@@@ Solid app set folder pane -- -- LOADING @@@@@')
```
dan-f: Is this supposed to stay here?

dan-f (approved):
👍 lgtm!

PR #18, opened 2017-11-21
Timbl

timbl:
This fixes the problem of the green plus on the databowser adding a phantom extra folder like "jkHSDFHksdhf   Test" 
and other stuff

### DIFF
--- a/contact/toolsPane.js
+++ b/contact/toolsPane.js
@@ -1,19 +1,16 @@
 //  The tools pane is for managing and debugging and maintaining solid contacts databases
 //
+/* global confirm, alert */
+
 var UI = require('solid-ui')
-var mime = require('mime-types')
 
 // var toolsPane
 var toolsPane = function (selectAllGroups, selectedGroups, groupsMainTable, book, dom, me) {
   var kb = UI.store
-  var ns = UI.ns
-  var updater = UI.store.updater
-  var ACL = UI.ns.acl
-  var VCARD = UI.ns.vcard
-  var doc = $rdf.sym(book.uri.split('#')[0]) // The ACL is actually to the doc describing the thing
+  const ns = UI.ns
+  const VCARD = ns.vcard
 
-  var buttonStyle = 'font-size: 100%; margin: 0.8em; padding:0.5em;'
-
+  const buttonStyle = 'font-size: 100%; margin: 0.8em; padding:0.5em;'
   var pane = dom.createElement('div')
   var table = pane.appendChild(dom.createElement('table'))
   table.setAttribute('style', 'font-size:120%; margin: 1em; border: 0.1em #ccc ;')
@@ -150,6 +147,7 @@ var toolsPane = function (selectAllGroups, selectedGroups, groupsMainTable, book
         //
         var eraseOne = function(card){
           return new Promise(function (resolve, reject) {
+
             var removeFromMainIndex = function() {
               var indexBit = kb.connectedStatements(card, stats.nameEmailIndex)
               log('Bits of the name index file:' + indexBit)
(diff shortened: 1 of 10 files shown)


contact/toolsPane.js
```diff
@@ -164,46 +162,47 @@ var toolsPane = function (selectAllGroups, selectedGroups, groupsMainTable, book
...
+              return resolve('Cancelled by user')
+            }
 
-          var deleteOneFile = function() {
-            var resource = filesToDelete.shift()
-            if (!resource) {
-              log('All deleted')
-              removeFromMainIndex()
-              return
+            var deleteNextFile = function () {
```
RubenVerborgh: Nit, but `function deleteNextFile ()` for such situations might be more idiomatic.

transaction/pane.js
```diff
@@ -198,15 +197,15 @@ module.exports = {
         var c3 = tr.appendChild(dom.createElement('td'))
         var amount = kb.any(x, ns.qu('in_USD'))
         c3.textContent = amount ? d2(amount.value) : '???'
-        c3.setAttribute('style', 'width: 6em; text-align: right; '); // @@ decimal alignment?
+        c3.setAttribute('style', 'width: 6em; text-align: right; ') // @@ decimal alignment?
```
RubenVerborgh, line 200: Perhaps a CSS class?

transaction/pane.js
```diff
@@ -220,37 +219,26 @@ module.exports = {
...
-        }
-        div.setAttribute('style', mystyle)
-      }
-
       var account = kb.any(subject, UI.ns.qu('toAccount'))
       setPaneStyle(account)
-      if (account == undefined) {
-        complain('(Error: There is no bank account known for this transaction <'
-          + subject.uri + '>,\n -- every transaction needs one.)')
+      if (!account) {
```
RubenVerborgh, line 227: Why not `=== undefined` here (as below)?

transaction/pane.js
```diff
@@ -453,20 +437,21 @@ module.exports = {
...
               }
             }
-
           })
 
           var types = []
           var grandTotal = 0.0
-          var years = [], i
+          var years = []
+          var i
```
RubenVerborgh, line 445: Also make this a `let` (as above)?

RubenVerborgh (commented):
Looks good, just a few minor nits.

contact/toolsPane.js
timbl: They are different, aren't they?   function x()  is hoisted up to th top of the containing function.  I thought.

RubenVerborgh: Yes, but since you're not referring to `deleteNextFile` above this line, they're the same.

transaction/pane.js
timbl: Using classes to trigger CSS style is document-wide.  You can't inject style sheets i a subtree.  So CSS classes  is a bit of  global file space, and so I avoid it.  You can't really use it with libraries.

contact/toolsPane.js
timbl: OK

timbl:
On the basis that this is better than master branch, and because use of npm link seems to be broke, and so that to do development one must push every small change through NPM instead of git, I think best to merge this.  (( like 
```
ERROR in ../solid-app-set/node_modules/auth-header/dist/index.js
    Module build failed: Error: Couldn't find preset "metalab" relative to directory "/devel/github.com/linkeddata/solid-app-set/node_modules/auth-header"
```
 and many more))

RubenVerborgh:
Maybe we can have a look at fixing the `npm link` scenario? I'm sure there's a way.

PR #20, opened 2018-01-20
Fix bugs in meeting scheduler; change some buttons from text to icons…

timbl:
… to be more international

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
   "dependencies": {
     "better-simple-slideshow": "linkeddata/better-simple-slideshow#gh-pages",
     "mime-types": "^2.1.13",
-    "solid-ui": ">=0.8.0"
-  },
+   "solid-ui": "file://../../linkeddata/solid-ui"
+   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
     "babel-core": "^6.24.1",
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
   "dependencies": {
     "better-simple-slideshow": "linkeddata/better-simple-slideshow#gh-pages",
     "mime-types": "^2.1.13",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -37,8 +37,8 @@
   "dependencies": {
     "better-simple-slideshow": "linkeddata/better-simple-slideshow#gh-pages",
     "mime-types": "^2.1.13",
-    "solid-ui": ">=0.8.0"
-  },
+   "solid-ui": "file://../../linkeddata/solid-ui"
```
melvincarvalho: should this say file?

schedule/schedulePane.js
```diff
@@ -537,19 +545,21 @@ module.exports = {
...
+      // emailButton.textContent = 'email invitations'
       emailButton.addEventListener('click', function (e) {
-        var title = '' + (kb.any(subject, ns.dc('title')) || '')
+        var title = kb.anyValue(subject, ns.cal('summary')) ||
+          kb.anyValue(subject, ns.dc('title')) || ''
         var mailto = 'mailto:' +
           kb.each(subject, ns.sched('invitee')).map(function (who) {
             var mbox = kb.any(who, ns.foaf('mbox'))
-            return mbox ? '' + mbox : ''
+            return mbox ? mbox.uri.replace('mailto:', '') : ''
```
melvincarvalho, line 558: could mbox be null here?  (dont know the answer, just checking)

timbl: Yes, in which case the empty string will be thing added to the main string.

timbl: Well, at the moment the form asks for a set of email addresses.  but later if it uses a people picker it coud  end up with a set of people soek of which don't have email addresses in the  data.

PR #24, opened 2018-02-24
Panes

timbl:
Refactor cleaner dependency on solid-ui
and separate out registry otherwise little change


timbl (commented):
Looks reasonable to me but I made it.. probably a light review by someone else a good idea

PR #25, opened 2018-03-07
Source code panes

timbl:
Add a source code view/edit pane.
Available for any document of type text/* or *xml* 
Allows editing in the browser and write back


melvincarvalho (commented):
Looks good!

I've tested this branch, and it works for me.

No obvious points in the code.  

One question I have is whether it's desirable to edit an LDPC and where such an edit would be saved?

PR #29, opened 2018-04-29
Timbl bugfix - fix problem of source pane not getting content type of document

timbl:
Had to switch to being able to take from previous HTTP transactions, as (a) the webOperation function doesn't save metadata in the store and (b) the response object in Chrome in this situation has response.headers as empty {}

### DIFF
--- a/source/sourcePane.js
+++ b/source/sourcePane.js
@@ -121,7 +121,7 @@ const thisPane = {
         setEditable()
       })
       .catch(function (err) {
-        div.appendChild(UI.utils.errorMessageBlock(dom, 'Error saving back: ' + err))
+        div.appendChild(UI.widgets.errorMessageBlock(dom, 'Error saving back: ' + err))
       })
     }
 
--- a/source/sourcePane.js
+++ b/source/sourcePane.js
@@ -146,13 +146,13 @@ const thisPane = {
         var contentType, allowed, eTag
         if (response.headers && response.headers.get('content-type')) {
           contentType = response.headers.get('content-type') // Should work but headers may be empty
-          allow = response.headers.get('allow')
+          allowed = response.headers.get('allow')
           eTag = response.headers.get('etag')
         }
 
-        let reqs = kb.each(null , kb.sym('http://www.w3.org/2007/ont/link#requestedURI'), subject.uri)
-        reqs.forEach( req => {
-          let rrr = kb.any(req , kb.sym('http://www.w3.org/2007/ont/link#response'))
+        let reqs = kb.each(null, kb.sym('http://www.w3.org/2007/ont/link#requestedURI'), subject.uri)
+        reqs.forEach(req => {
+          let rrr = kb.any(req, kb.sym('http://www.w3.org/2007/ont/link#response'))
           if (rrr) {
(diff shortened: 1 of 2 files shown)


source/sourcePane.js
```diff
@@ -151,14 +143,23 @@ const thisPane = {
...
-        let rrr = kb.any(response.req, kb.sym('http://www.w3.org/2007/ont/link#response'))
-        if (rrr) {
-          contentType = kb.anyValue(rrr, UI.ns.httph('content-type'))
-          allowed = kb.anyValue(rrr, UI.ns.httph('allow'))
-          eTag = kb.anyValue(rrr, UI.ns.httph('etag'))
-          if (!eTag) console.log('sourcePane: No eTag on GET')
+        var contentType, allowed, eTag
+        if (response.headers && response.headers.get('content-type')) {
+          contentType = response.headers.get('content-type') // Should work but headers may be empty
+          allow = response.headers.get('allow')
```
melvincarvalho: Possibly define allow above

melvincarvalho (changes requested):

melvincarvalho:
When hitting the saveBack button I get : 

```
Uncaught Error: Web operation sending data must have a defined contentType.
    at Fetcher.webOperation (mashlib.js:50631)
    at HTMLButtonElement.saveBack (mashlib.js:14860)
```

source/sourcePane.js
```diff
@@ -151,14 +143,23 @@ const thisPane = {
...
         setUnedited()
-        var contentType, allowed
-        let rrr = kb.any(response.req, kb.sym('http://www.w3.org/2007/ont/link#response'))
-        if (rrr) {
-          contentType = kb.anyValue(rrr, UI.ns.httph('content-type'))
-          allowed = kb.anyValue(rrr, UI.ns.httph('allow'))
-          eTag = kb.anyValue(rrr, UI.ns.httph('etag'))
-          if (!eTag) console.log('sourcePane: No eTag on GET')
+        var contentType, allowed, eTag
+        if (response.headers && response.headers.get('content-type')) {
```
melvincarvalho: contentType is already defined as a var in render, should it also be a var in refresh, or inherit -- unsure

melvincarvalho:
I think `UI.utils` -> `UI.widgets` here : 

https://github.com/solid/solid-panes/blob/4727bb3a9dbe0265850d5f033060567d194ef6a7/source/sourcePane.js#L123

source/sourcePane.js
timbl: Good point ...

timbl: should be 'allowed'

timbl: Fixed

source/sourcePane.js
timbl: Mybad.  refresh() should inherit the outer level render() ones

melvincarvalho (approved):

melvincarvalho:
This works locally for me.  I'm happy to merge if you are!

timbl:
@melvincarvalho  Thanks for you review -- it really needed it!  It works for me now on localhost.  Works for you too?

timbl:
Ok Thanks!!

PR #35, opened 2018-07-05
Fix typos in README

maael:
Just fixing a few typos in the README. Had to guess a bit with the `Do anyting`, but I think I got it right.


melvincarvalho (approved):

melvincarvalho:
@maael Thanks so much!

Issue #34, opened 2018-05-28
Error: Cannot mint new - missing newInstance


tmciver:
# Summary

When attempting to create a new address book, the error `TypeError: Cannot read property 'uri' of null` is shown.

---

# Steps to Reproduce
1. Click "Public Folder" from your account home page on solid.community.
2. Click the green plus sign.
3. Select "Make new Address Book".
4. Enter a name in the text box.
5. Click the green check mark.

Error is displayed in browser.

melvincarvalho:
When I try this now, I get no error, but also no address book!  Curious!

tmciver:
That is strange.  I'll see if I can have a look tomorrow.

Vinnl:
This bug was fixed in d356d18f2c88ae21899b4636aaca74aed0873627, and the address book project has since moved to https://github.com/solid/contacts-pane.

PR #38, opened 2018-08-29
Fix for bug #34.

tmciver:
Fix for the "Error: Cannot mint new - missing newInstance" error that happens when attempting to create a new address book.

### DIFF
--- a/contact/contactPane.js
+++ b/contact/contactPane.js
@@ -120,7 +120,8 @@ module.exports = {
           p.innerHTML =
             "Your <a href='" + uri + "'><b>new " + appInstanceNoun + '</b></a> is ready. ' +
             "<br/><br/><a href='" + uri + "'>Go to new " + appInstanceNoun + '</a>'
-          resolve(context)
+          var newContext = Object.assign({ newInstance: kb.sym(uri) }, context)
+          resolve(newContext)
         }
 
         var doNextTask = function () {


contact/contactPane.js
```diff
@@ -120,7 +120,8 @@ module.exports = {
           p.innerHTML =
             "Your <a href='" + uri + "'><b>new " + appInstanceNoun + '</b></a> is ready. ' +
             "<br/><br/><a href='" + uri + "'>Go to new " + appInstanceNoun + '</a>'
-          resolve(context)
+          var newContext = Object.assign({ newInstance: kb.sym(uri) }, context)
```
melvincarvalho, line 123: Thanks for the contribution.  Looks like a simple enough change!

Did you have a chance to test it yet?

tmciver: @melvincarvalho I tested it manually in the browser but did not write any programmatic tests (I don't see any existing tests in the code base).

tmciver: @melvincarvalho Did you want me to add some sort of programmatic tests or is manual testing enough?

melvincarvalho: Apologies for the delay in response!

Tests are always a plus, but not necessary in this case.  Just testing it locally should be fine.  I did discuss this with @timbl but he's on vacation right now.

The question he had was whether the Object assign should have context first or 2nd.

Im not that familiar with the code, but try and understand it more.

tmciver: It does not matter what order the objects are in in `Object.assign`.  I went off of this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign and followed their example of putting the literal first and variable second.

FYI: I have a refactor of some of this code in the works but I wanted to get this bug fix merged first.

melvincarvalho: OK, I think it looks good.  Only concern was this part here : 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Deep_Clone

It does a copy by reference, rather than, a deep clone.  But I think you are minting a new reference, so that should be OK?

If you're good with this, I'm tempted to go ahead and merge ...

tmciver: You're right: it only does a shallow copy so it's not giving the immutability I was going for (and neither does the similar "spread" operator `...`).  This is probably fine since there is probably mutation happening elsewhere in the application.

Any interest in using [Immutable.js](https://facebook.github.io/immutable-js/)?

melvincarvalho: That could be ok.  I saved a comment from @timbl 

> The pattern with the context object is to accumulate the data on a common onbject, so this needs to be the other way around.

See what you think, whether we should swap?

tmciver: Doing `Object.assign(context, { newInstance: kb.sym(uri) })` mutates `context`.  It would probably be clearer to simply do `context.newInstance = kb.sym(uri)` but I was trying to avoid mutation.

melvincarvalho: Good point!

It looks to be called from here : 

https://github.com/solid/solid-ui/blob/5c8165af0155fb9ba3033ae88b006f531738b8f5/src/create.js#L75

When the promise returns it's looking for the .newInstance -- so maybe this isnt working at all.

> context.newInstance = kb.sym(uri)

or the Object.assign I think should work, but as I say I'm very unfamiliar with this code.

I'm tempted to merge, then have @timbl look over it when he gets back, on the basis that it seems unlikely to break anything further, and can fix at least one bug.  What do you think?

tmciver: Yes, [the error on line 78](https://github.com/solid/solid-ui/blob/5c8165af0155fb9ba3033ae88b006f531738b8f5/src/create.js#L78) is exactly what this change is fixing.  I think it's fine to merge; we can always revisit later.

melvincarvalho: But I think as per timbls comment the context should go first, preserving the reference, or add the kb.sym(uri) to the object -- then we've also taken into account the feedback.  Does that make sense?

tmciver: Oh, I thought you meant you'd merge it as is then get @timbl's feedback later.  Do you want to go with the mutating version? Personally, I avoid mutation in my code but if you want me to change it, I will.  I don't believe there's any need for the mutation - at least I would hope that none of the code relies on it, but I don't know for sure.

melvincarvalho:
@timbl we chatted about this and I passed on your feedback.  We think it's good to merge and will fix a bug.  Would be great if you had feedback, when you are around!

timbl:
The question is with the makeNew API when a context is passed the original one is expected to be altered, or whether new versions will only be available in the context passed on in the promise close.  I think the latter is better practice (allows immutable objects to be used) but the question isn't documented at the moment. With a typed language, we could declare the parameter read-only.

PR #39, opened 2018-09-27
Use foaf:name in profile editor

RubenVerborgh:
When preparing the demo for the Decentralized Web Summit, we noticed that:
- a fresh user's profile on node-solid-server uses `foaf:name`
- the profile editor of solid-panes uses `vcard:fn`

The result is that the profile editor will not show the person's name.

### Change the profile editor to use `foaf:name` instead _(= this PR)_
Problem: FOAF does not have all properties that VCARD has, so we end up with a mixture.

### Change node-solid-server to default to `vcard:fn`
Problem: FOAF is popular, it seems to be assumed in the footprint of a FOAF profile

### Make the editor accept both `vcard:fn` and `foaf:name`
Problem: I don't know how to model this with `http://www.w3.org/ns/ui#`, maybe this isn't possible yet.

### Change node-solid-server to default to use both `vcard:fn` and `foaf:name`
Problem: redundancy / inconsistencies?

### Use reasoning to derive one property from the other
That's what I'm doing to publish my profile (https://ruben.verborgh.org/articles/queryable-research-data/#reason).
Problem: I think this is the way forward for the future (query in one ontology and have reasoning derive the rest), but is likely too complex for now.

### DIFF
--- a/contact/individualForm.js
+++ b/contact/individualForm.js
@@ -1,6 +1,7 @@
 module.exports = `
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
 @prefix ui: <http://www.w3.org/ns/ui#>.
+@prefix foaf: <http://xmlns.com/foaf/0.1/>.
 @prefix : <#>.
 
 <http://www.w3.org/2006/vcard/ns#Individual>
@@ -22,9 +23,8 @@ module.exports = `
 
     <#fullNameField>
         a <http://www.w3.org/ns/ui#SingleLineTextField> ;
-        ui:label "Name";
         <http://www.w3.org/ns/ui#maxLength> "128" ;
-        <http://www.w3.org/ns/ui#property> <http://www.w3.org/2006/vcard/ns#fn> ;
+        <http://www.w3.org/ns/ui#property> foaf:name ;
         <http://www.w3.org/ns/ui#size> "40" .
 
     <#roleField>
--- a/contact/individualForm.ttl
+++ b/contact/individualForm.ttl
@@ -1,5 +1,6 @@
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
 @prefix ui: <http://www.w3.org/ns/ui#>.
+@prefix foaf: <http://xmlns.com/foaf/0.1/>.
 @prefix : <#>.
 
 <http://www.w3.org/2006/vcard/ns#Individual>
@@ -21,9 +22,8 @@
 
     <#fullNameField>
         a <http://www.w3.org/ns/ui#SingleLineTextField> ;
-        ui:label "Name";
         <http://www.w3.org/ns/ui#maxLength> "128" ;
-        <http://www.w3.org/ns/ui#property> <http://www.w3.org/2006/vcard/ns#fn> ;
+        <http://www.w3.org/ns/ui#property> foaf:name ;
         <http://www.w3.org/ns/ui#size> "40" .
 
     <#roleField>


melvincarvalho:
You didnt give a link to the profile editor, there's more than one.

This is what I've done in apps in the past : 

```JavaScript
    var FACE  = $rdf.Namespace("https://graph.facebook.com/schema/~/");
    var FOAF  = $rdf.Namespace("http://xmlns.com/foaf/0.1/");

    g.any(kb.sym(webid), FOAF('name'))  || g.any(kb.sym(webid), FACE('name'))
```

I'm happy to add vcard to this list.  Lately I've been using vcard groups (foaf does them too).

So with this code, it can use either the foaf name or the facebook name (which I had integrated).  I'm sure there will be more in future such as AS, Schema etc.

melvincarvalho:
Ah I see, it's in the pull request.  OK that's for creating a profile.  My code above was for consuming.  There possibly might be some docs in the solid tree for standard profile fields.  I'll send them to you in chat if I find them.

melvincarvalho:
Found it!

https://github.com/solid/solid-spec/blob/master/solid-webid-profiles.md

timbl:
A very good question. I am inclined not to change the contact form to use a mixture.  Moving toward vCard for everything makes more sense I think. Contact interop if going to be key.  That said,  enhancing the form system so that it can populate a form from legacy properties but save them back as he preferred ones might be generally useful. Should that be put in the form or maybe even driven by ontology level data like that one predicate superseeds another

timbl:
(This is stuff we can do with forms. Meanwhile a useful function which devs may find useful is one which for any object returns the best thing we can figure out as a name for it.  One place where we put a bunch of smarts including guessing from a profile uri before loading it even. )

RubenVerborgh:
> I am inclined not to change the contact form to use a mixture.

Agree on that.

> Moving toward vCard for everything makes more sense I think.

I think it's a matter of context.

Within a contact card or address book, vCard makes most sense.
For this profile editor, where 98% of fields will be vCard anyway, it doesn't feel right to have one property as FOAF—especially since that exact property is also covered by vCard.

However, there are other contexts where FOAF makes sense, and one of those context is (obviously) a FOAF profile. As @melvincarvalho points out, this is the choice we made in the Solid WebID Profiles Spec.

Both choices are logical: vCard in a contact card context, FOAF in a profile context. Since both choices are logical, none is necessarily better than the other; it's about fitness for purpose.

My long-term bet is on leveraging the semantics in the data and going from one to the other on-the-fly thanks to reasoning. It's the best way around it, and I have this on my long-term roadmap.

But we also need to make choices for the shorter term.

> Contact interop if going to be key.

True, but that doesn't mean that we have to make cross-app agreements on everything—or we lose some of the benefits of decentralization. It's good to have a small set of things we absolutely agree on, but there should be optional sets that you can opt in or out of.

It also doesn't mean that we have to make these choices at the source of the data.

In the W3C Data Exchange Working Group, I'm working on profile-based content negotiation. I hope, in the future, to be able to extend the current mechanism of "the client prefers Turtle" with "the client prefers Turtle where people are represented with FOAF and publications with BIBO". Then the server can perform reasoning and format translation on the fly and give every client what they need. (To be solved: writing.)

> That said, enhancing the form system so that it can populate a form from legacy properties

I would be hesitant to call FOAF "legacy". There are many use cases for FOAF, also for vCard, and also for Schema.org.

We need a generation of applications that can deal with different ontologies; otherwise we might very well skip RDF and just agree on JSON key names instead of property URIs. The fact that we have semantics in our data is almost equally important as the fact that we have links, so we should put those semantics to use.

JordanShurmer:
> We need a generation of applications that can deal with different ontologies; otherwise we might very well skip RDF and just agree on JSON key names instead of property URIs. The fact that we have semantics in our data is almost equally important as the fact that we have links, so we should put those semantics to use.

Great point, well said.

timbl (changes requested):
This is a much longer discussion which we have since had..  I thnk having a mixture in the vcard card is noit a way to go by the way

PR #41, opened 2018-10-03
Allow iframe to request camera and mic for video conf in meeting iframe

timbl:
We are iframe-ing jit.si but that blocks camera permission in current chrome.  See

 https://dev.chromium.org/Home/chromium-security/deprecating-permissions-in-cross-origin-iframes


meeting/meetingPane.js
```diff
@@ -805,6 +805,8 @@ module.exports = {
         iframe.setAttribute('src', target.uri)
         // iframe.setAttribute('style', 'height: 350px; border: 0; margin: 0; padding: 0; resize:both; width: 100%;')
```
pfriedman, line 806: While we are in this section of the file might as well remove the old commented out style. Not particularly necessary but calling out the [boy scout rule](https://medium.com/@biratkirat/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385)

pfriedman (approved):
This looks like a fine change

PR #49, opened 2018-10-24
Need to ignore the IDE resources Jetbrains use

megoth:
Just a small MR to get things going

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ node_modules
 
 # Optional REPL history
 .node_repl_history
+
+# IDE resources
+.idea


RubenVerborgh:
This should be in your own global .gitignore, no?

megoth:
Could do that as well, but why not have it in the repo for others that use Jetbrains IDEs as well? Instead of relying on everyone having this as part of their global gitignore (I honestly didn't know of global gitignore before your comment =P )

RubenVerborgh:
> but why not have it in the repo for others that use Jetbrains IDEs as well?

To avoid all the PRs from people using 37 other editors ;-)

megoth:
> To avoid all the PRs from people using 37 other editors ;-)

Hah, true; we could take a copy from https://www.gitignore.io/ in that case (they usually covers all IDEs and stuff), but will maybe do that later.

Ok, just ignore this commit, I'll do this from my global gitignore.

PR #50, opened 2018-10-24
Toggle to uneditable state upon saving a source

megoth:


melvincarvalho:
@megoth nice touch!  Did you test it?

Should we have a test URI where we can try out such changes -- would be awesome methinks ...

megoth:
> @megoth nice touch! Did you test it?

Yes, I tested it out using node-solid-server ^_^

> Should we have a test URI where we can try out such changes -- would be awesome methinks ...

Even better would be some sort of unit tests, but it's difficult the way the code is structured now. We might get some testing [using Snapshots (e.g. via Jest)](https://jestjs.io/docs/en/snapshot-testing).

timbl:
This edit has been overtaken by events, in that it was scooped up at the same time as the UI was revised when the syntax check was put in

PR #52, opened 2018-11-11
Source pane syntax check

timbl:
Checking the syntax before saving a data file is just an obvious thing to do.  We can all make mistakes. 

Also, users can end up editing data file by accident  and ending up in a catch-22 situation with an illegible ACL file which the system prevents them from ever changing again.

When the user presses the checkmark button then there is a syntax check.
This syntax check displays an error in a status box and colors the user text red.   Because we don't have instant syntax coloring (would be nice!!) , the moment the text is touched it goes green (save me color) even before they have actually got the syntax right.

Also now the save button is only visible when the text has been changed, and so should be saved.  When a successful save happens, it disappears.  This also is a necessary feedback that the sabe has happened at all, which was too obscure before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,12 +34,20 @@
     "url": "https://github.com/solid/solid-panes/issues"
   },
   "homepage": "https://github.com/solid/solid-panes",
+  "dependenciesFile": {
+    "@solid/better-simple-slideshow": "^0.1.0",
+    "babel-preset-env": "^1.6.1",
+    "babel-preset-metalab": "^1.0.0",
+    "mime-types": "^2.1.13",
+    "rdflib": "file://../../linkeddata/rdflib.js",
+    "solid-ui": "file://../../solid/solid-ui"
+  },
   "dependencies": {
     "@solid/better-simple-slideshow": "^0.1.0",
     "babel-preset-env": "^1.6.1",
     "babel-preset-metalab": "^1.0.0",
     "mime-types": "^2.1.13",
-    "rdflib": ">=0.17.0",
+    "rdflib": ">=0.17.1",
     "solid-ui": ">=0.11.5"
   },
   "dependenciesBackup": {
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -34,29 +34,29 @@
...
   "homepage": "https://github.com/solid/solid-panes",
-  "dependencies": {
+  "dependenciesFile": {
     "@solid/better-simple-slideshow": "^0.1.0",
     "babel-preset-env": "^1.6.1",
     "babel-preset-metalab": "^1.0.0",
     "mime-types": "^2.1.13",
-    "rdflib": ">=0.17.0",
-    "solid-ui": ">=0.11.5"
+    "rdflib": "file://../../linkeddata/rdflib.js",
```
megoth, line 42: Please revert this

megoth: That is, I'm guessing this kind of fixes to make it work locally shouldn't be here?

megoth (commented):
Apart from the `depenciesFile`-field in package.json, I think this looks good. I'm not certain if my request is correct, as I'm not familiar with `depenciesFile` in general. Is it just a handy way for easily testing this repo locally, not affecting other developers? If so, just ignore my comments.

But I want to test it locally before I approve it, to be on the safe side. I'm sitting on a bus now with low battery, but will try to get it reviewed when I'm back home.

RubenVerborgh (approved):
LGTM!

RubenVerborgh:
We might want the server to check this for RDF files. We check on PATCH, but not yet PUT or POST. It doesn't make a lot of sense to PUT/POST a syntactically invalid RDF file.

megoth:
> We might want the server to check this for RDF files. We check on PATCH, but not yet PUT or POST. It doesn't make a lot of sense to PUT/POST a syntactically invalid RDF file.

That was my plan with https://github.com/solid/node-solid-server/issues/882, i.e. a simple validation that basicly makes sure that RDF-data to be stored are sound. And then there's https://github.com/solid/node-solid-server/issues/907 which discusses a broader set of validation.

package.json
timbl: The dependenciesFile and dependenciesBackup are just places to keep different versions - because JSON doesn't allow comments :-(   
I find them handy and some other people find them a useful tip. but you can't test them

timbl:
I'll merge this noting that the error messages work but are suboptimal because the error objects made by the parser need to be standardized for line number and character number in file .. that's a task in rdflib.js at some point  --> https://github.com/linkeddata/rdflib.js/issues/275

megoth:
Finally got to test this locally; it works as intended. But I'm critical of the handling of the Save-button - I've created an issue for this https://github.com/solid/solid-panes/issues/53.

PR #54, opened 2018-11-15
Remove unused external JS links and update prefixes

csarven:
Minor update. No change to functionality.

### DIFF
--- a/dokieli/new.html
+++ b/dokieli/new.html
@@ -1,23 +1,20 @@
 <!DOCTYPE html>
 <html lang="en" xml:lang="en" xmlns="http://www.w3.org/1999/xhtml">
-    <head>
-        <meta charset="utf-8" />
-        <title></title>
-        <meta content="width=device-width, initial-scale=1" name="viewport" />
-        <link href="https://dokie.li/media/css/basic.css" media="all" rel="stylesheet" title="Basic" />
-        <link href="https://dokie.li/media/css/do.css" media="all" rel="stylesheet" />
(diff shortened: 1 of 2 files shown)


megoth (approved):
LGTM

melvincarvalho:
Im sure you're aware, but just in case.  

Changes to this library need to be propagated upstream to dependencies (mashlib and node-solid-server) via npm as well.

RubenVerborgh:
> Changes to this library need to be propagated upstream to dependencies (mashlib and node-solid-server) via npm as well.

That's not correct and it's important we get this right.

As long as the dependency is within the semver range of those dependents, then that version will be used automatically on installs.

You _might_ want to change the package-lock.json of dependents (and you can decide individually which ones) such that, as a _developer_ checking out the repository, you use those latest versions. But again, non-developers doing a package installation from npm will get the latest dependencies.

csarven:
So, is there something you'd like me to do further for this PR? AFAICT, the current changes in this PR doesn't require an update to package*.json. If package*.json is currently behind, that's another matter.

RubenVerborgh:
The PR is fine.

melvincarvalho:
> That's not correct and it's important we get this right.

@RubenVerborgh Agree.  Perhaps we could add some notes somewhere, eg in an .md file.  

I think we're saying the same thing.  

In either case, the version of npm will be need to be bumped for this package.

That will then propagate to dependencies (libraries, and ultimately, pods) either semi-automatically (with a minor version) or manually (for a patch version) depending which is deemed appropriate.  

In this case there are fewer round trips, and better privacy, but functionality should remain backwards compatible, so could be either.

PR #56, opened 2018-11-19
Modular panes

timbl:
Refactor the NPM around sold-panes pulling out initially 6 panes to be separate NPM modules so they can have their own issue trackers etc


package.json
```diff
@@ -46,9 +46,16 @@
...
+    "chat-pane": "^1.2.3",
+    "contacts-pane": "^1.0.3",
+    "folder-pane": "^1.0.4",
+    "issue-pane": "^1.0.3",
+    "meeting-pane": "^1.0.2",
     "mime-types": "^2.1.13",
+    "pane-registry": "^1.0.4",
     "rdflib": ">=0.17.1",
-    "solid-ui": ">=0.11.5"
+    "solid-ui": ">=0.11.5",
```
RubenVerborgh, line 57: We might want to practice semver for this dependency too.
Are there still (m)any dependencies on solid-ui left?

RubenVerborgh (approved):
LGTM

package.json
timbl: Well every pane depends individually on solid-ui

RubenVerborgh: Yes, but does solid-panes directly?

timbl: Semver with version 0 considered harmful ... unexpected inconsistencies deemed between minor versions causes unneeded duplicates, which with solid-auth-client lead to strange bugs.. inclined to keep semver caret until we get to 1.0.0

RubenVerborgh: Definitely. I was rather suggesting we bring solid-panes up to 1.0.0.

PR #59, opened 2019-03-03
An attempt at a new pane - trustedApplications

### LINKED ISSUE #58: Make it possible for users to manage trusted apps
With the implementation of `acl:trustedApp` in [acl-check](https://github.com/solid/acl-check/issues/17) and [node-solid-server](https://github.com/solid/node-solid-server/issues/1121), it would be nice if we offered a UI to allow users to manage their trusted apps.

I propose to do this as an extension of [the profile pane](https://github.com/solid/solid-panes/blob/master/profile/profilePane.js).

megoth:
Allows you to manage web applications by granting (or removing) their origin access.

Should fix https://github.com/solid/solid-panes/issues/58.

Implementations for acl-check and node-solid-server are at https://github.com/solid/acl-check/pull/18 and https://github.com/solid/node-solid-server/pull/1122 respectively.

### DIFF
--- a/index.js
+++ b/index.js
@@ -130,6 +130,7 @@ register(require('./internalPane.js'))
 // The home pane is a 2016 experiment. Always there.
 
 register(require('./profile/profilePane.js')) // edit your public profile
+register(require('./trustedApplications/trustedApplicationsPane.js')) // manage your trusted applications
 register(require('./home/homePane.js'))
 
 // ENDS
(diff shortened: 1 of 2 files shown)


michielbdejong (approved):
I tried it out and it seemed to work correctly.
We would need this merged before Friday ideally, because it's a dependency of the NSS 5.0 release)

timbl:
If you tried it out and it works then that's good for me.    A new pane is new functionality which is very unlikely to break other things, so in general should be encouraged even if they may be imperfect.

timbl (commented):
LGTM

PR #60, opened 2019-03-10
new dokieli bits, more VisualLanguage.html, bump solid-ui and chat-pane

timbl:


csarven:
I think https://github.com/solid/solid-panes/pull/57 should be merged first and this PR base off that.

melvincarvalho:
> I think #57 should be merged first and this PR base off that.

Had a quick look at this change.  Quick question : 

Is there not a slight privacy issue with solid users being tracked by -- dokie.li

     <script src="https://dokie.li/scripts/dokieli.js"></script>

Pretty sure that site is high trust, and this look OK.  

But as we tend to err on the side of caution, wrt the privacy surface.  Just wondering if this is a known issue, or there's workarounds I am not that familiar with.

timbl:
@csarven taken which version of the files, from #57 or #60 ?

timbl:
The long name versions of course.. not sure how to do that in the github UI

timbl:
Merged with the longer names dokieli.js etc

timbl:
Yes @melvincarvalho at the moment we have to trust dokie.li and many places from which we get load out code... should we move them all to github instead?

melvincarvalho:
> Yes @melvincarvalho at the moment we have to trust dokie.li and many places from which we get load out code... should we move them all to github instead?

Good question!

We should probably make a list of the external scripts and css.  Then try to reduce the surface.

For example.  Let's say someone got access to dokieli and changed the script to download people's inbox.  We would not have an easy way to detect that, instead of say if it was inline on github.

You can make the same argument about github I suppose.  Which is an issue in itself.  Just thinking we should possibly document the privacy / attack surface, and also ways that we can mitigate that.  Maybe some of these things are non issues already, or could be with certain counter measures.

csarven:
Just need to have a position on whether solid-panes should rely on any script and media from different origins. Should solid-panes be self-contained? Should certain operations be carried out strictly on localhost or offline?

So, solid-panes having its own copy of dokieli's JS and CSS probably takes care of those concerns with the trade-off of using dokie.li as a CDN (type of thing).

Aside: I'd be more than happy to take additional measures to improve whatever is needed on the dokie.li end to minimise potential issues. Please let me know!

legendarymaster31:
I think we should make another different separate pane for this purpose.

megoth:
Maybe Subresource Integrity is a way of verifying the script loaded from dokie.li? https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

megoth:
As long as https://dokie.li/scripts/dokieli.js don't change often, that could work - but maybe it does?

PR #66, opened 2019-04-03
Shapes documentation

michielbdejong:
I started collecting some examples of data shapes that the various panes of the databrowser store on your pod.

Will remove the '[WiP]' prefix from the title when this is ready for review.


michielbdejong:
See also https://github.com/labra/solid-shapes/issues/1 and https://github.com/solid/chat-pane/pull/15.

Sorry I haven't had time to complete this yet! Will try to give it higher priority this week.

michielbdejong:
Feedback from @pmcb55: There's a slight typo on that 'conventions.md' page - you have 'sioc' listed twice (I forked the repo, but can't see your branch to make a PR to fix this myself). Secondly, both DC and DCTERMS vocabs are being used here (which is confusing and outdated, see: https://stackoverflow.com/questions/47519315/what-is-the-difference-between-dublin-core-terms-and-dublin-core-elements-vocabu). I guess we need to be backward compatible, so NSS 4 and 5 will need to continue to support both, but I think we should mark DC usage as deprecated and recommend DCTERMS exclusively from now on. (Also I think we should be using the prefix 'dcterms:' (instead of 'dct:'), as that's more readable, and also the prefix they use themselves, see: curl -L -H "Accept: text/turtle" http://purl.org/dc/terms/). Finally(!), neither DC or DCTERMS seem to define an ':author' property at all (although it's mentioned in multiple places in your document), so that looks like a bug to me (but one we'll probably have to keep for backward compatibility again!). I think it should be replaced with 'dcterms:creator'.

Documentation/conventions.md
```diff
@@ -0,0 +1,235 @@
...
+* @prefix foaf: <http://xmlns.com/foaf/0.1/>.
+* @prefix ical: <http://www.w3.org/2002/12/cal/ical#>. 
+* @prefix ldp: <http://www.w3.org/ns/ldp#>.
+* @prefix mee: <http://www.w3.org/ns/pim/meeting#>.
+* @prefix pim: <http://www.w3.org/ns/pim/space#>.
+* @prefix rdf: http://www.w3.org/2000/01/rdf-schema#>.
+* @prefix schema: <http://schema.org/>.
+* @prefix sioc: <http://rdfs.org/sioc/ns#>.
+* @prefix solid: <http://www.w3.org/ns/solid/terms#>.
+* @prefix sioc: <http://rdfs.org/sioc/ns#>.
```
TallTed: line 23 duplicates line 21

Issue #63, opened 2019-03-20
profile grows when adding trusted apps


michielbdejong:
On gitter, @angelo-v [wrote](https://gitter.im/solid/node-solid-server?at=5c90f5a0f3dbbd230c808a23):
> Just tried the "trusted app" pane in NSS 5. My profile grows from 2.9K to 267K after doing so and contains tons of triples that should not be there... Anyone experienced this already?
Just seems that the whole knowledge graph has been PUT there by the client

angelo-v:
Version: solid-server@5.0.0 from npm

## How to reproduce?

1. Go to your solid profile page (logged in)
2. Open the "Manage your trusted applications pane"
3. Enter an application URL, e.g. "https://markbook.org"
4. Select desired Access modes (I chose Read, Write, Append, but do not know if it is relevant)
5.  Click the add button

## Expected behaviour
A triple is added to the solid profile, stating that the app is trusted

## Actual behaviour
Countless, non-related and non-desired triples are added to the profile. (excerpt below). This leads to a approx. factor 10 growth of the file size of card$.ttl.

## Excerpt of unwanted triples

```
ind:fullNameField
    a ui:SingleLineTextField;
    ui:label "Name";
    ui:maxLength "128";
    ui:property n:fn;
    ui:size "40".
ind:fullNameFieldC
    a ui:SingleLineTextField;
    ui:maxLength "128";
    ui:property n:organization-name;
    ui:size "40".
ind:id1409437207443
    a ui:SingleLineTextField;
    ui:maxLength "128";
    ui:property n:street-address;
    ui:size "40".
ind:id1409437292400
    a ui:SingleLineTextField;
    ui:maxLength "128";
    ui:property n:locality;
    ui:size "40".
ind:id1409437421996
    a ui:SingleLineTextField;
    ui:maxLength "25";
    ui:property n:postal-code;
    ui:size "25".
ind:id1409437467649
    a ui:SingleLineTextField;
    ui:maxLength "128";
    ui:property n:region;
    ui:size "40".
ind:id1409437569420
    a ui:SingleLineTextField;
    ui:maxLength "128";
    ui:property n:country-name;
    ui:size "40".
ind:id1409437646712 a ui:Classifier; ui:from rdf:Class; ui:property terms:type.

ind:noteComment a ui:Comment; ui:contents "General Notes".
```

bourgeoa:
NSS v5.0.0
[...]

PR #68, opened 2019-04-25
Changed save logic to make use of updateManager.update

megoth:
This sends SPARQL UPDATEs that should fix the problem mentioned in https://github.com/solid/solid-panes/issues/63

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,6 @@ node_modules
 
 # Optional REPL history
 .node_repl_history
+
+# files used by IDEs such as WebStorm
+.idea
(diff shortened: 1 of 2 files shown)


angelo-v:
If it "should fix" #63 there should be a test that proves that and ensures that it never occurs again.

Vinnl (approved):

Vinnl:
I added tests for it in #77.

megoth:
This is a fix outside of the current work on revamp, so will merge it into master and then rebase revamp on master afterwards

Issue #69, opened 2019-04-25
Introduce TypeScript


megoth:
As part of the work on the Solid data browser (that we mentioned in https://github.com/solid/solid-panes/issues/64 earlier) we want to introduce TypeScript to improve the code quality and making future refactoring of the code base easier. We will work on this in a separate branch until we know that it is stable enough and have consensus from the community.

[TypeScript](https://www.typescriptlang.org/) offers a variety of features that makes working with JavaScript easier. We suggest migrating code file by file as we improve the codebase, and making sure to enhance the new files with types where needed (ie set `noImplicitAny` to `true`).

We need to make sure that the build setup is rigid but flexible enough so that developers who do not want to use TypeScript for their work can opt out.

gobengo:
@megoth In my experience a really good way to start introducing the type-checking benefits (e.g. for refactoring like you mention) without taking the potentially controversial (but IMO good) step of actually moving to the superset of the language is to take advantage of the fact that `tsc` can type-check JS files based on JSDoc comments.

I have used it in the past to add type-safety at time of continuous testing, but in a project where the owner wasn't sure about shifting languages. It worked quite well and caught many bugs that the team would have otherwise shipped to prod.

https://github.com/Microsoft/TypeScript/wiki/JSDoc-support-in-JavaScript

A slow-adoption with this as a milestone could also get merged back into master sooner and be a smaller [batch size](https://medium.com/the-agile-weekly/understanding-product-development-batch-size-work-in-process-wip-risk-for-small-teams-f551144ecbb5) of work, rather than maintaining a separate branch indefinitely, which can spiral out of control.

megoth:
@gobengo thank you for your feedback. It is a good proposal, but we opted for moving files to TypeScript.

The code has been merged into the branch https://github.com/solid/solid-panes/tree/revamp, but we hope to merge that into master soon. We just need to resolve some questions with Tim first.

PR #74, opened 2019-05-07
Add Webpack config

Vinnl:
There were references to Webpack and Babel in the code, but it appears that the code was published as-is anyway.

I've now added Webpack configuration and set up Babel. I tried to replicate the workings of the current code with the built code, and it appears to run fine locally. However, I don't know how the code is usually published, and I don't have that much experience with the data browser yet that I can really stress test all edge cases. In other words: please do feel free to extensively point out how this could be better :)

This could pave the way to fixing #69.


Vinnl:
Oops, forgot to actually activate Babel... One moment >.<

Vinnl:
Sorry about that, fixed now.

megoth (approved):
LGTM

megoth:
Merging this into branch revamp until we have things a bit more in order.

PR #75, opened 2019-05-07
Activate TypeScript

### LINKED ISSUE #69: Introduce TypeScript

Vinnl:
(Note: only relevant after #74 is merged.)

This PR enables type checking. The approach is as follows:

- There's a Babel plugin to remove type annotations from the code.
- In parallel to compiling the code, Fork-TS-Checker-Plugin (for Webpack) checks the type annotations.

I chose this approach because it appears we're already using Babel, and this makes it easy not to have type checking block compilation, which helps with gradual adoption.

TypeScript is currently set to strict mode, but is only active for `.ts` files. This means we can gradually migrate by renaming files, and simply disable type checking for specific values using `any` when it's too much work to introduce proper types.

Critique away! :)

Fixes #69.


megoth (approved):
In general I want to avoid type `unknown` as much as possible, but will allow it for this PR.

Vinnl:
@megoth They're basically there to indicate "do not forget to properly type this later". Since they cannot be used until they get properly typed anyway, I figured this is a way to avoid typing values that are currently unused (hence speeding up the migration), especially considering that I do not know what the proper interface would be :) (See also the `PaneDefinition` interface, which I think is probably mostly correct, but is still the result of me guessing the interface based on other code and hence might not be entirely correct.)

In other words: I'd be happy for you to insert the correct types if you know them, but given that an incorrect type is probably worse than no type, I left it at this :)

PR #76, opened 2019-05-08
Add testing infrastructure

### LINKED ISSUE #73: Introduce testing
We need to find a way to introduce tests in the current codebase. This will probably involve finding some sensible ways of separating some code into specific concerns, to figure out what is sensible to test and what is not (e.g. certain DOM manipulations might not be sensible to test).

This work item is heavily related to the work item [Implement conventions that facilitate separation of concern for panes](https://github.com/solid/solid-panes/issues/72), and one should not be done without considering the effects it will have on the other.

As pointed out in https://github.com/solid/solid-panes/pull/68#issuecomment-489916216 we should have a test for the trustedApp logic to make sure that there is no regression on the functionality. I think this is a good pane to start writing tests for.

Vinnl:
Unfortunately, this also depends on the Webpack config from #74, because Jest appears to expect Babel - and a recent version at that.

This PR adds a basic testing setup, and an example test. It is currently limited to logic that does not interact with the DOM and can be extracted into its own file. We could theoretically use something like [Karma](https://karma-runner.github.io/) to run tests in the browser and test actual DOM interactions, but since those would be a lot more brittle and are often not the most critical pieces of code to test, I figured this was fine for now.

Let me know what you think.

Fixes #73.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2454,18 +2454,6 @@
         "text-encoding": "^0.6.1"
       }
     },
-<<<<<<< HEAD
-    "@types/rdflib": {
-      "version": "0.17.1",
-      "resolved": "https://registry.npmjs.org/@types/rdflib/-/rdflib-0.17.1.tgz",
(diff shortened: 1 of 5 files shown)


megoth (approved):
LGTM

PR #84, opened 2019-05-13
Convert TrustedApplicationsPane (and tests) to TypeScript

Vinnl:
To make sure that an actual pane could be written in TypeScript.

One thing I encountered that's good to know: Babel does not support mixing CommonJS modules and ES modules, so make sure to use `export default` and `import {} from` instead of `module.exports =` and `require()` when using TS.


trustedApplications/trustedApplicationsPane.ts
```diff
@@ -92,28 +109,41 @@ function createApplicationTable(subject) {
     createText('th', 'Access modes'),
     createText('th', 'Actions')
   ])
-  applicationsTable.appendChild(header)
+  applicationsTable.appendChild(header);
```
megoth, line 112: Minor comment: We need to avoid minor diffs in code style such as this (some places we have semi-colon, others not). We've used StandardJS for this before, lets make sure that the same rules apply for our [TypeScript](https://standardjs.com/rules.html) files (or something close enough, at least)

trustedApplications/trustedApplicationsPane.ts
```diff
@@ -92,28 +109,41 @@ function createApplicationTable(subject) {
...
+    .forEach(({appModes, origin}: {appModes: NamedNode[], origin: NamedNode}) => applicationsTable.appendChild(createApplicationEntry(subject, origin, appModes, updateTable)))
 
   // adding a row for new applications
   applicationsTable.appendChild(createApplicationEntry(subject, null, [ns.acl('Read')], updateTable))
 
   return applicationsTable
 
   function updateTable() {
-    applicationsTable.parentElement.replaceChild(createApplicationTable(subject), applicationsTable)
+    applicationsTable.parentElement!.replaceChild(createApplicationTable(subject), applicationsTable)
```
megoth, line 129: Nice, didn't know of [the non-null assertion operator in TS](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator), always fun to learn new stuff ^_^

megoth (approved):
Found some faulty assumptions in one of the tests, and we need to set up a linter for our TypeScript files, but all in all a good fix.

trustedApplications/trustedApplicationsPane.ts
Vinnl: I tried to respect the current coding style. In this case, however, the semi-colon can not be removed, because the following statement is enclosed in brackets for the type cast. Otherwise it would interpret that statement as an argument passed to a function that `appendChild` should have removed.

That said, when I ran StandardJS on the JS files, there were quite a lot of errors...

trustedApplications/trustedApplicationsPane.ts
Vinnl: :) I prefer not to use it, but since it worked without a non-null check so far, I figured it'd be OK to keep it that way.

trustedApplications/trustedApplicationsPane.ts
megoth: Yes, we need a strategy for how to resolve this when migrating JS files.

PR #85, opened 2019-05-13
Do not modify the History on direct links

### LINKED ISSUE #1094: Home page WebID link breaks browser back button
I'm using a local development deployment of solid-server, as described for #1093.

Starting point is  that the public homepage (https://localhost:8443/) is displayed in my browser (currently using Firefox for this test).

If I click on the "WebID" link (https://localhost:8443/profile/card#me), a new page is displayed at the link location, but I can't click the browser's "back" button to return to the home page - nothing happens.

Vinnl:
When arriving at a page through a direct link, that page is already
added to the browser's History. Thus, if we added it through
history.pushState again, it'll be duplicate, and the browser's Back
button no longer works.

Fixes https://github.com/solid/node-solid-server/issues/1094.

### DIFF
--- a/outline/manager.js
+++ b/outline/manager.js
@@ -1662,7 +1662,12 @@ module.exports = function (doc) {
       UI.utils.getEyeFocus(tr, false, undefined, window) // instantly: false
     }
 
-    if (solo && dom && dom.defaultView && dom.defaultView.history) {
+    if (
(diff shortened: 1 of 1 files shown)


megoth (approved):
Tested and verified on my local setup ^_^

PR #87, opened 2019-05-14
Introduce linting to TypeScript

### LINKED ISSUE #86: Introduce linting to our TypeScript code
We should introduce linting for TypeScript to ensure a common code style.

We are using [Standard JS](https://standardjs.com/) to lint our JavaScript code today, and this is a library used in many other Solid projects (meaning projects living in the Solid GitHub organization). Lets see if there are any linters out there with a configuration setup that are similar to Standard JS, so that we might have more harmony between our JS and TS code.

Vinnl:
As before, this uses [Javascript "Standard" Style](https://standardjs.com/). That said, it only applies it to the TypeScript files, as turning it on for Javascript resulted in 6003 errors and 997 warnings, even though I have not touched most of the code.

Fixes #86.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,3 @@
+{
+  "extends": "eslint-config-standard"
+}
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,22 @@
+module.exports = {
+  'env': {
+    'browser': true,
+    'es6': true,
+    'node': true
+  },
+  'extends': 'standard',
+  'globals': {
+    'Atomics': 'readonly',
+    'SharedArrayBuffer': 'readonly'
+  },
+  'parser': '@typescript-eslint/parser',
+  'parserOptions': {
+    'project': './tsconfig.json',
+  },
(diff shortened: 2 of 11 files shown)


megoth (approved):
Added a config file so that IDEs such as WebStorm are able to enforce the linting rules. Apart from that it looks very good ^_^

megoth:
As you mention, there are files that won't pass the linting if we lint JS-files as well, so I think we need to do some pre-linting commits when we migrate JS-files to TS in the future. Lets discuss this more tomorrow.

Vinnl:
@megoth There already was a config file! It was `.eslintrc.js`. If WebStorm doesn't support that we could convert it to `.eslintrc`, but I think there's a few things missing in the file you committed...

megoth:
Oh, sorry, I missed that (trying to do a review in between sessions wasn't the smartest move =p ) - then it's probably just a question of merging the configs, or discarding one of them (probably mine).

I'll give it a go later this evening, unless you beat me to it ^_^

PR #97, opened 2019-05-24
Autogenerate API docs from JSDoc comments

### LINKED ISSUE #96: Generate API documentation for TypeScript files

Vinnl:
This adds the command `npm run docs` to generate API documentation. It doesn't appear to be that useful currently, given that solid-panes is not really a library to be used by other projects, but give it a look and see what you think.

I'm not sure if and where we'd want to publish this, so I did not include CI configuration to generate the docs and deploy it somewhere. This might be something we'd want to look at later.

Fixes #96.


megoth:
True, we want to expose API documentation for the parts that will be exposed to others - this might be limited to how the data browser will expose APIs to panes, as that is probably the interface that other developers will need to care about.

Lets put this on hold for now and pick it back up when we know more of how the codebase will evolve in the coming weeks.

megoth:
Also, just to note: I was hoping typedoc could reuse the types used in signatures (in TypeScript files), so that we don't need to be redundant.

megoth:
I'll close this PR for now, since we have https://github.com/solid/solid-panes/issues/96 as a reminder for this later.

PR #99, opened 2019-05-27
Proposal for dynamic panes

### LINKED ISSUE #72: Implement conventions for panes that facilitate separation of concern
Some panes are difficult to understand today for various reasons: Some files are huge (close to 1000 SLOC), some functions are dependent on specific scoping, and some are very much dependent on intricate ways of manipulating the DOM.

We suggest introducing patterns/techniques and tools that are reused across the various panes to make it easier to get an overview of what a pane does.

A technology that might be of interest here is Web Components, that might allow us to create reusable components that can be reused across panes, perhaps even in other projects. A tool that might be of interest here is Stencil, which might make it easier for us to create web components (which could be easily ported to major frameworks as well).

*Action items*

- Find a pattern that faciliates a sensible separation of concern for panes
- Make the pattern a convention across all panes
- Optional: Make use of Web Components
- Optional: Make use of Stencil

Vinnl:
(This PR builds on #100, and is thus easier to review if that is merged first.)

This Pull Request contains a new Pane ("scratchpad", similar to pad) whose only goal is to demonstrate the concepts. It is intended as a starting point for a discussion of how best to implement dynamic loading of panes, and how to structure a Pane's code. In other words: feel free to liberally criticise anything in here :) 

It proposes that:

- Pane definitions no longer rely on global variables like `window.UI`, because a developer can then not reliably trace where a piece of code came from. Instead, e.g. store instances are passed as parameters - which is why this depends on https://github.com/solid/solid-ui/pull/64, and why `outline/manager.js` was modified.
- Panes are defined in separate Node packages, even though they can share some common ceremonial code (e.g. Webpack config). This allows them to be distributed independently from the data browser, without us having to reinvent a package manager and all its associated problems.
- Code to manipulate data in the store (i.e. `data.ts`) is decoupled from the Pane definition (although it's currently in the same package). This makes it easier for third parties to create apps that manipulate the same data in the same way. It could also allow for e.g. a dashboard-like replacement for the current HTML home page of the data browser, which can re-use that code to e.g. initialise new resources that can be read by that pane.
- Not relying on the `label` method to determine whether a pane should be shown, to lower the barrier to entry for new developers.
- The view is still somewhat cumbersome, but at least it's tested without firing up a fickle browser (using jsdom).
- A wrapper connecting the proposed pane definitions to the data browser's current methods. This enables the new pane definitions to avoid relying on global data directly, and thereby potentially be included more easily in e.g. a new data browser, a single-pane shell (i.e. pane-to-app converter), or a [storybook](https://storybook.js.org/).

Open challenges:

- The pod server should expose to the data browser which panes are installed and hence available to use - at least as a first step; a later step could be discussing loading panes from arbitrary locations, as that comes with its own set of challenges.
- How much freedom do we want to give to panes? In the current implementation, a pane receives a DOM element that it can fill with whatever they want. Pane authors could decide to use React, for example, which means that if two panes get loaded that both use React, they both instantiate their own render trees. It also means that if, e.g. a user switches away from a pane and back again, that pane cannot efficiently preserve and restore its state.

An overview of what's what in this PR, to hopefully make it easier to review:

- ~~Stricter ESLint and test coverage setup (for TypeScript files).~~ Moved to #100.
- **`packages/panes`**: this is where separate Node packages (i.e. with their own `package.json`) for panes could live.
- **`packages/initialise.js`** (and other scripts it references): since `npm run watch` only compiles and starts `solid-panes`, this script does the same for all panes in separate packages inside `packages/panes`.
- **`view.ts` and `data.ts`**: view and data access for a pane (scratchpad, in this case). And their test files.
- **paneWrapper.ts`**: interface between the old pane definition and the new one.
- **`packages/panes/webpack-config-panes`**: the only non-pane inside `packages/panes`. This exports a method that can be used to initialise the Webpack config for a pane in a separate package. Currently only used by the scratchpad, but allows different panes to re-use the configuration if needed.
- `pad/padPane.ts`: a rewrite of the regular pad Pane from Javascript to TypeScript, no functional changes. Mostly done to be able to compare scratchpad.

Apologies for the enormous code drop, but given that it's quite a task to move to dynamic panes, I'm afraid there's just a lot to discuss. As said, nothing here is set in stone, as far as I care.

Fixes #72.


Vinnl:
I just extracted a few unrelated commits out into #100, so it can more easily be reviewed/merged in pieces. Apologies if you already pulled this branch locally. If so, and you did not make changes yet locally, please run:

    git checkout revamp; git branch --delete 72-scratchpad; git fetch; git checkout 72-scratchpad

types.ts
```diff
@@ -1,11 +1,41 @@
-import { Node } from 'rdflib'
+import { IndexedFormula, NamedNode } from 'rdflib'
 
 // Note: there might be a more appropriate project to hold this definition:
 export interface PaneDefinition {
   icon: string;
   name: string;
-  label: (subject: Node) => string | null;
-  render: (subject: Node, dom: HTMLDocument, options?: unknown) => HTMLElement;
```
Vinnl, line 8: First feedback by @megoth: move `store` to be part of `options` in `render` and `mintNew`. (Not sure if there's a better solution for `label`?)

Reference: https://github.com/solid/solid-ui/pull/64#pullrequestreview-242711412

megoth:
For others reading this - this is a proposal connected to https://github.com/solid/solid-panes/issues/103

megoth:
As noted in [this comment](https://github.com/solid/solid-panes/issues/103#issuecomment-496614851) we'll postpone this work until we have more work done on other parts of the data browser.

I do think there are some code in this PR that we'll merge in as separate PRs, but for now I'll close this PR.

PR #100, opened 2019-05-28
Refinement of test setup

Vinnl:
This adds:

- ESLint permission to prefix unused variable with an underscore. This is not allowed by default by Standard JS, but since a lot of current code does not work with named parameters, and we sometimes need the third but not the second argument to a function, being able to indicate that the second variable is unused is sometimes necessary.
- Sets test coverage thresholds to 100%, only applying to TypeScript (i.e. new) files. Note that not all lines actually have to be covered - see [here](https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md) for instructions on how to indicate that you decided not to add tests for specific lines of code.
- Run Jest in watch mode by default, allowing you to keep it running while working on tests, and automatically re-executing when you change the tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,8 @@
     "watch": "npm run clean && webpack --progress --colors --watch --mode=\"development\"",
     "clean": "rm -rf dist/",
     "lint": "eslint \"**/*.ts\"",
-    "test": "npm run lint && jest --onlyChanged --watch"
+    "test-watch": "npm run lint && jest --onlyChanged --watch",
+    "test": "npm run lint && jest"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "watch": "npm run clean && webpack --progress --colors --watch --mode=\"development\"",
     "clean": "rm -rf dist/",
     "lint": "eslint \"**/*.ts\"",
-    "test": "npm run lint; jest"
+    "test": "npm run lint && jest --onlyChanged --watch"
```
megoth: I think we should have a separate script for watch, as `npm test` should be a script that ends (this will keep running, watching for changes). Maybe `test-watch` to be aligned with other scripts in this project?

megoth (changes requested):
Add another script, then this should be good.

package.json
Vinnl: :+1:

PR #105, opened 2019-05-29
Fork rdflib typings to fix gaps in the definitions

Vinnl:
Submitted upstream at
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35783

Once that is merged, we can pull the new version and this can be reverted.


megoth (approved):
LGTM

PR #107, opened 2019-05-29
Proposal for refactoring

Vinnl:
(Builds on #106.)

This introduces a new Pad pane that can read Pads created by
the old one. It is intended as a testbed for experimenting with a
new setup for Panes in which their concerns are more clearly
separated. Specifically, it aims to achieve the following goals:

- Extract data access into a separate module. This potentially
  allows these modules to be re-used in other apps, to ensure
  data compatibility.
- Separate DOM manipulation from data manipulation. This makes it
  easier to update the view without risk of breaking something
  else.
- Add a wrapper to intermediate between the old and the new API.
  Hopefully we can let go of the old API in time, but that first
  requires more clarity on what needs to be preserved.
- Separate creation of new items from viewing them. Since the
  data model can be shared between different apps/panes, no single
  pane should be responsible for creating new instances, as that
  would be redundant with the equivalent methods in different
  panes.
- Separate item viewing from item discovery. Panes are responsible
  for viewing specific pieces of data (in this case: a notepad).
  This is useful for e.g. when a user is linked to a resource
  directly, allowing them to browse a specific piece of data.
  Discovering what data is available on one's Pod in general is a
  separate task - one could imagine e.g. a listing of simply all
  notepads in a Pod, sorted by date. Clicking one of those would
  bring you to the Scratchpad Pane.

This API is not final.

Note that it makes use of async/await, and therefore imports
@babel/polyfill. Because mashlib overwrites `require`, we cannot
yet include it globally, so the import has to be added to every
file using ES modules and async/await.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+typings/rdflib
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,5 @@ install:
   - npm install
 script:
   - npm test
+  # Make sure it still builds successfully - the build isn't actually used yet:
+  - npm run build
--- a/index.js
+++ b/index.js
@@ -51,6 +51,7 @@ register(require('issue-pane'))
 register(require('contacts-pane'))
 
 register(require('./pad/padPane').default)
+register(require('./scratchpad/paneWrapper').default)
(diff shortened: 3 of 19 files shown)


megoth (approved):
LGTM

PR #116, opened 2019-06-11
[Work in progress] Separation of concerns using Surplus

megoth:
Trying out new conventions using [Surplus](https://github.com/adamhaile/surplus)

This library allows us to write markup with JSX and handles internal state.

I've split code into controllers, views, models, and services.

This PR is to showcase the changes I've done, should not be merged before a broader discussion.


typings/marked/index.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'marked' {
```
Vinnl, line 1: Where [the DT definitions](https://www.npmjs.com/package/@types/marked) insufficient?

Vinnl, line 1: Where [the DT definitions](https://www.npmjs.com/package/@types/marked) insufficient?

types.ts
```diff
@@ -32,24 +33,10 @@ export interface ViewParams {
...
   user?: NamedNode;
-};
+}
 
 export interface RevampPaneDefinition {
   canHandle: (subject: NamedNode, store: IndexedFormula) => boolean;
   view: (params: ViewParams) => void;
   label: (subject: NamedNode, store: IndexedFormula) => string | null;
-};
-interface NewPaneOptions {
```
Vinnl, line 42: Ehm... Did I add this twice? How did that happen? >.<

typings/solid-ui/index.d.ts
```diff
@@ -1,6 +1,14 @@
+
 declare module 'solid-ui' {
-  export type SolidUi = any
+  export type SolidUi = {
```
Vinnl, line 3: This PR should make this file redundant: https://github.com/solid/solid-ui/pull/70

webpack.config.js
```diff
@@ -10,10 +10,11 @@ module.exports = {
     libraryTarget: 'commonjs2',
   },
   resolve: {
-    extensions: ['.mjs', '.js', '.ts'],
+    extensions: ['.mjs', '.js', '.ts', '.tsx']
   },
   module: {
     rules: [
+      { test: /\.tsx$/, loader: 'surplus-loader!ts-loader' },
```
Vinnl, line 17: [Babel can load TypeScript files too](https://vincenttunru.com/why-combine-babel-and-typescript/), so I don't think there's a need to introduce ts-loader here.

package.json
```diff
@@ -88,7 +94,9 @@
     "eslint-plugin-standard": "^4.0.0",
     "fork-ts-checker-webpack-plugin": "^1.3.0",
     "jest": "^24.8.0",
+    "surplus-loader": "^0.5.0",
     "ts-jest": "^24.0.2",
+    "ts-loader": "^6.0.2",
```
Vinnl, line 99: As said below, I think ts-loader shouldn't be necessary.

trustedApplications/trustedApplications.controller.ts
```diff
@@ -0,0 +1,97 @@
+import { S } from 'surplus'
+import { DataSignal } from 's-js/src/S'
+
+import solidUi from 'solid-ui'
+import $rdf, { NamedNode } from 'rdflib'
+import { TrustedApplication } from './trustedApplications.models'
+import { getStatementsToAdd, getStatementsToDelete } from './trustedApplications.service'
+
+const { authn, ns, store } = solidUi
```
Vinnl, line 9: I think it'd be good to move away from the SolidUI Singleton, insofar as possible, i.e. by limiting access to it to `trustedApplications/trustedApplicationsPane`, and have that pass the required data (e.g. the current session) as a parameter to other modules that need it.

trustedApplications/trustedApplications.models.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.isNew = !_origin
+    this.origin = S.value(_origin)
+    this.modes = ['Read', 'Write', 'Append', 'Control'].map(mode => ({
+      name: mode,
+      value: ns.acl(mode),
+      isChecked: S.value(_modes.some(appMode => appMode === ns.acl(mode).value))
+    }))
+  }
+
+  static createNew (_subject: NamedNode): TrustedApplication {
```
Vinnl, line 24: I'm wondering whether a Factory might be overengineering this. As an alternative, we could also just define an interface using something like this:

```typescript
export interface TrustedApplication {
    origin: DataSignal<string>
    subect: DataSignal<string>
    modes: Mode[]
}

export const fromSubject: (subject: NamedNode) => TrustedApplication = /* ... */
export const fromNamedNodes (subject: NamedNode, origin: NamedNode | null, modes: NamedNode[]) => TrustedApplication = /* ... */
```

Then no private constructor is needed, nor `copy` (which would just be `const copiedApp = {...originalApp}`), and a simple object would suffice.

trustedApplications/trustedApplications.view.tsx
```diff
@@ -0,0 +1,74 @@
+import * as Surplus from 'surplus'
+import { TrustedApplicationsController } from './trustedApplications.controller'
+import solidUi from 'solid-ui'
+import data from 'surplus-mixin-data'
+import { TrustedApplication } from './trustedApplications.models'
+
+const { S } = Surplus
+const { widgets } = solidUi
+
+export const TrustedApplicationsView = (controller: TrustedApplicationsController) =>
```
Vinnl, line 10: It is not directly clear to me that the separation of the view and the controller offers much value here - they seem pretty tightly coupled, with the controller's methods being closely related to what the view is doing. Would it not be more readable to have that logic co-located with the view that is influenced by it?

Vinnl (commented):
Overall I really like the direction this is going in! It's a clear improvement over what we have now, in my eyes. I've added some food for thought inline that we can discuss.

The one major point, which I'm guessing you're probably already expecting, concerns the addition of Surplus. While I generally like its concepts, it comes with clear longevity concerns, which means we should be prepared to fork and maintain it eventually - and I'm not sure whether we have the capacity to do that, and whether that'd be worth it.

Thus, I think the primary thing I'd like a stronger rationale for, is: why not go with e.g. React, especially considering that that would allow us to reuse and dogfood the SDK?

typings/marked/index.d.ts
megoth: Oh, there's a definition already, didn't see that - should make use of that instead.

trustedApplications/trustedApplications.controller.ts
megoth: I'm thinking along this lines as well, but didn't want to revolutionize the code base all at once =P Probably need a migration path for this, but could of course introduce that in these conventions

megoth:

megoth:
We've decided to use [React for this purpose going forward](https://github.com/solid/solid-panes/pull/120).

PR #117, opened 2019-06-12
[Work in progress] Separation of concerns using React

### LINKED ISSUE #72: Implement conventions for panes that facilitate separation of concern

Vinnl:
This is strongly based on #116. I created a Markdown pane with practically the same functionality and a similar architecture as in that PR, but this time using React. This primarily demonstrates that it is possible to use React in a single pane, without having to share a data model with the rest - it simply attaches to a DOM node, which is then provided to the OutlineManager.

The primary change is that there is no controller here. That said, there still is a "Container" component, which glues the data fetching to the view (ie hard to test, but logic that's not too complicated). That means that the view logic is relatively easy to test - an example is included.

I'll look at porting the TrustedApplications Pane next to get a feel for how this would work for a more involved Pane, but thought I'd submit this to give an initial view of what this might look like.

Fixes #72.


markdown/view.test.tsx
```diff
@@ -0,0 +1,44 @@
+/* eslint-env jest */
+import * as React from 'react'
+import {
+  render,
+  fireEvent
+} from '@testing-library/react'
+import { View } from './view'
+import { workaroundActError } from './actErrorWorkaround'
+
+workaroundActError()
```
megoth, line 10: Very nice that we can do snapshot testing!

markdown/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+        newInstance
+      }))
+      .catch((err: any) => {
+        console.error('Error creating new instance of markdown file', err)
+        return options
+      })
+  },
+  render: (subject) => {
+    const container = document.createElement('div')
+    ReactDOM.render(<Container store={store} subject={subject}/>, container)
```
megoth, line 29: Very simple connection between panes and React... very interesting

markdown/container.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react'
+import { loadMarkdown, saveMarkdown } from './service'
+import { View } from './view'
+import { ContainerProps } from '../types'
+
+export const Container: React.FC<ContainerProps> = (props) => {
```
megoth, line 6: Props should be typed of course, but looks like a reasonable way of passing data

markdown/view.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const [rawText, setRawText] = React.useState(props.markdown)
+
+  function storeMarkdown () {
+    setPhase('saving')
+    props.onSave(rawText).then(() => {
+      setPhase('rendering')
+    })
+  }
+
+  if (phase === 'saving') {
```
megoth, line 20: I like this approach better than the conditionals in my experiment - I think this is easier to read.

megoth (commented):
A lot of improvements here, and I'm more positive to React with this. Lets try a bit more experimenting and see where they take us.

markdown/container.tsx
Vinnl: They are typed :) The generic passed to `React.FC` (i.e. `ContainerProps`) defines the types of the props. (I chose to extract them to a separate file here to encourage using the same API for all top-level containers - that would make it easier to potentially extract panes into separate apps, for example.)

megoth:

megoth:
Should keep this open until we've merged it into master.

angelo-v:
@Vinnl I was going to work on a markdown pane and just found this! Any reason we should not merge it?

Vinnl:
@angelo-v IIRC this was to prototype the use of React in solid-panes, which is quite a dependency. And the other reason would be the large number of conflicts :P

angelo-v:
We included react already. My activtiystreams-pane is also react-based. So this should not be an issue. 
I guess the conflicts should also not be too hard, since it is quite well seperated

Vinnl:
In that case I see no reason not to merge it :)

angelo-v:
Ok, I will give it a try

angelo-v:
I moved the code to a dedicated repo https://github.com/solid/markdown-pane

PR #120, opened 2019-06-13
[Work in progress] Separation of concerns using React - TrustedApplications

Vinnl:
I've separated this out from #117 to highlight only what is new compared to that. Like that PR, this is basically replicating the approach from #116 using React.


trustedApplications/service.ts
```diff
@@ -0,0 +1,76 @@
...
+        })
+    })
+    .sort((appA: TrustedApplication, appB: TrustedApplication) => (appA.origin > appB.origin) ? 1 : -1)
+}
+
+/**
+ * @param serialisedMode The full IRI of a mode
+ * @returns A plain text string representing that mode, i.e. 'read', 'append', 'write' or 'control'
+ */
+export function deserialiseMode (serialisedMode: $rdf.NamedNode, ns: Namespaces): Mode {
```
Vinnl, line 68: @megoth Do you know if there's a built-in way to remove the namespace from an IRI?

Vinnl:
Btw, this is what it now looks like, given that the style guide is active now:

image

trustedApplications/__snapshots__/service.test.ts.snap
```diff
@@ -43,7 +43,7 @@ Array [
   Statement {
     "object": NamedNode {
       "termType": "NamedNode",
-      "value": "http://www.w3.org/ns/auth/acl#Read",
+      "value": "http://www.w3.org/ns/auth/acl#read",
```
megoth, line 46: Nitpicking, but this should not be the case - Read (and the other modes) are classes, so should start with capital letters

trustedApplications/container.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return <div>Your profile {props.subject.doc().uri} is not editable, so we cannot do much here.</div>
+  }
+
+  const fetchedTrustedApps: TrustedApplication[] = fetchTrustedApps(props.store, props.subject, ns)
+
+  const [trustedApps, setTrustedApps] = React.useState(fetchedTrustedApps)
+
+  const addOrEditApp = (origin: string, modes: Mode[]) => {
+    const result = new Promise<void>((resolve) => {
+      const deletions = getStatementsToDelete($rdf.sym(origin), props.subject, props.store, ns)
```
megoth, line 27: I'm not a big fan of this level of scoping =/ (This is where I like to rely on classes instead, which of course introduces some challenges of its own.) Just wanted to mention it.

trustedApplications/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+import { PaneDefinition } from '../types'
+import { Container } from './container'
+
+const nodeMode = (typeof module !== 'undefined')
+
+let panes
+let UI: SolidUi
+
+if (nodeMode) {
```
megoth, line 24: Know that we have to do this wrt testing, but also would like a better approach at some point.

trustedApplications/service.ts
megoth: You mean something like `https://www.w3.org/ns/auth/acl#Read` - namespace = `Read`?

trustedApplications/view.tsx
```diff
@@ -0,0 +1,215 @@
...
+        </form>
+      </td>
+      <td>
+        <button onClick={() => props.onDelete(props.app.origin)}>Delete</button>
+      </td>
+    </tr>
+  )
+}
+
+const NewApplication: React.FC<{ onSave: AddOrUpdate }> = (props) => {
```
megoth, line 139: Hmm, perhaps better conceptually with two components, but adds a bit of overhead in terms of maintenance =/

trustedApplications/container.tsx
```diff
@@ -0,0 +1,81 @@
...
+  )
+}
+
+function insertTrustedApp (app: TrustedApplication, into: TrustedApplication[]): TrustedApplication[] {
+  const index = into.findIndex(found => found.origin === app.origin)
+  if (index === -1) {
+    return into.concat(app)
+  }
+
+  return into.slice(0, index)
```
megoth, line 68: Wouldn't it be better to use [Array.prototype.splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice)?

megoth (commented):
I'm liking the general direction this is going! Good use of container-view pattern. It's not perfect, but I think it's a big step in the right direction.

I want to finish my Vue-experiment, and then we'll try to wrap this up.

trustedApplications/__snapshots__/service.test.ts.snap
Vinnl: That's exactly the kind of nitpicking I need, otherwise I'd never learn. Is a "class" an RDF concept, and if so, how do I know when something is a class?

(This is also probably wrong at many other places.)

trustedApplications/container.tsx
Vinnl: Me neither, didn't think too long about this. Ideally, the updater would return a Promise itself. (Well, there's actually code that does, but it's not quite clear to me yet when - might be an avenue to pursue. If we do go ahead with this, I'll see if I can improve this.)

trustedApplications/index.tsx
Vinnl: That's not there for testing - it's primarily there because it already was there; see [here](https://github.com/solid/solid-panes/blob/f2b7c78a1b669f9ba235b0ace1b02a91e40b9f49/trustedApplications/trustedApplicationsPane.js#L13).

I think it may have to do with having been a Firefox extension or something? Perhaps Tim knows more about it.

trustedApplications/service.ts
Vinnl: Yes, exactly.

trustedApplications/view.tsx
Vinnl: I thought it'd make it easier to read, but I'm certainly open to inlining it.

trustedApplications/container.tsx
Vinnl: I usually go for methods that do not modify their input parameters, but sure, `splice` would work too.

megoth:

trustedApplications/service.ts
megoth: Not that I'm aware of, no, but shouldn't be to difficult to create

megoth:
We will merge [soc-react](https://github.com/solid/solid-panes/tree/soc-react) and [soc-react-trustedApps](https://github.com/solid/solid-panes/tree/soc-react-trustedapps) into master.

PR #121, opened 2019-06-20
[Work in progress] Separation of concerns using Vue

megoth:
Trying out new conventions with Vue.

Using TypeScript and Templates, supporting Sass for styling.


megoth:

markdown/App.vue
```diff
@@ -0,0 +1,67 @@
...
+    components: {
+      MarkdownEdit,
+      MarkdownView
+    }
+  })
+  export default class App extends Vue {
+    state: STATE = STATE.LOADING
+    markdown: string = ''
+
+    @Prop(NamedNode) subject!: NamedNode
```
Vinnl, line 26: What does the exclamation mark do here? I know it as casting to a non-null/undefined value, but since this appears to be a property declaration, I presume that's not what's happening here?

markdown/App.vue
```diff
@@ -0,0 +1,67 @@
...
+  import { loadMarkdown, saveMarkdown, STATE } from './markdown.service'
+  import MarkdownEdit from './components/MarkdownEdit.vue'
+
+  @Component({
+    components: {
+      MarkdownEdit,
+      MarkdownView
+    }
+  })
+  export default class App extends Vue {
```
Vinnl, line 22: Not a biggy, but if we go this route, we really shouldn't start calling Panes Apps as well, because that's a surefire way to confuse everybody and annoy Daphne :P

markdown/components/MarkdownEdit.vue
```diff
@@ -0,0 +1,34 @@
...
+  export default class MarkdownEdit extends Vue {
+    @Prop() markdown!: string;
+    raw!: string
+
+    created () {
+      this.raw = this.markdown
+    }
+
+    save (): void {
+      this.$emit('save', this.raw)
```
Vinnl, line 23: I presume this is throwing an event, but I haven't found yet where it is caught - can we/TypeScript trace that?

(I've seen projects where it was unclear whether an event could be removed, because there was no good way to tell whether there were still listeners that were expecting it - or vice versa, that it was not clear whether a listener could be removed. Debugging was also hampered because stack traces ended at the event.)

trustedApplications/components/TrustedApplicationRow.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+    <tr>
+        <td>
+            <input class="textinput" placeholder="Write new URL here" v-model="application.origin" />
+        </td>
+        <td>
+            <label v-for="mode in application.modes">
+                <input type="checkbox" v-model="mode.isChecked"/>
```
Vinnl, line 8: Would TypeScript warn you if you'd have misspelled `isChecked` here? (Asking for this one since it appears to be the most complicated one, so if it works here that'd inspire confidence in the rest.)

Vinnl (commented):
Overall this'd work fine, I think. I'm not too big a fan of the need for so many custom build tooling and Vue-specific attributes, but it's a big enough framework I think that that shouldn't be _too much_ of a problem.

Do we have an idea of what the upgrade story to v3 is going to be like? One thing we should look out for is not only changes that break our code, but changes that break some of the tooling (bundles, editors, etc,), and that tooling not being able to keep up.

Speaking of which: how was WebStorm's support? I recall that VSCode has the vetur plugin which is supposedly pretty good, so it's interesting to see what that story is like for other editors. E.g. did you get syntax highlighting for Vue attributes in templates? Type completion?

markdown/App.vue
megoth: If I don't use the exclamation mark I have to write `NamedNode | undefined` - this would be more correct if I couldn't guarantee subject being set before the created event (where I need it), but I can guarantee it so thought it was better to use the exclamation mark.

markdown/components/MarkdownEdit.vue
megoth: It is caught by the `v-on:save="toggle"` that you'll find in `Markdown/App.vue`.

This is how [event handling are done in Vue](https://vuejs.org/v2/guide/events.html). I think it's fairly easy to follow the trace, as it's always direct connection between parent-child, not grandparent-child (not 100% sure on this though).

markdown/App.vue
Vinnl: Ah, I see - because you're declaring but not initialising the property, TS thinks it could be undefined. (And theoretically it can be, if you don't pass the prop in `index.ts`'s `render` function. But we do, and its type does not include `undefined` there; so yeah, it's justified.

trustedApplications/components/TrustedApplicationRow.vue
megoth: WebStorm gives me feedback that something is wrong by coloring `isChecked` differently based on if it's a known property or not. In addition, Vue will output errors in the console.

In general I find the error messages that Vue outputs quite specific and helpful.

Vinnl: Nice!

Vinnl: (Does it also indicate that `isChecked` is a boolean?)

megoth:
We've decided to use [React for this purpose going forward](https://github.com/solid/solid-panes/pull/120).

PR #123, opened 2019-06-27
Global nav and URL bar setting

### LINKED ISSUE #122: shift-click should update the URL in the address bar

timbl:
Goal: fix bug that URKL bar not always set when should be

add login/out button on righthand side
add  global nav menu when click on the login button a bit like github etc etc

also ... Remove UI.isExtension  code as that was for an obsolete Firefox extension

Fixes #122.

### DIFF
--- a/home/homePane.ts
+++ b/home/homePane.ts
@@ -15,6 +15,8 @@ import panes from 'pane-registry'
 const HomePane: PaneDefinition = {
   icon: UI.icons.iconBase + 'noun_547570.svg', // noun_25830
 
+  global: true,
+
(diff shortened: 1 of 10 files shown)


megoth (approved):
The amount of changes in this PR makes it nearly impossible to review properly. Most of the changes are done in outline/manager.js as well, which is excruciating to maintain because of its shear size and complexity.

That being said, I have managed to run this locally, and I trust Tim's insight into how these will affect the outcome.

(But we seriously need to improve our routines and testability if we want this code repo to be somewhat maintainable.)

PR #126, opened 2019-06-27
change fake.* to *.example in sample URIs

TallTed:


TallTed:
recreating PR against `solid:master`

PR #128, opened 2019-06-28
change fake.* to *.example in sample URIs 

TallTed:


TallTed:
If I understand the Travis report details correctly, the failure is being caused by URIs which are being constructed by appending `_line0`, `_line1` and the like to existing URIs (which end at authority, and do not include any PATH element at all) -- which is just plain wrong.

I *think* whatever code is appending these suffixes-with-leading-underscores should be adjusted to instead append suffixes-with-leading-octothorps, i.e., `#line0`, `#line1`, etc.

But I don't know enough about the code behind to fix that error.

Vinnl:
The failure was caused by not all `.snap` files being updated in line with the corresponding changes in the test files. I've bundled your commits together (30afaf4daa958be506b38f808b53c925c620be7f) and fixed the failures in 30afaf4daa958be506b38f808b53c925c620be7f.

Thanks for your contribution!

PR #130, opened 2019-07-16
Add the dashboard from datasister as a Pane

Vinnl:
We can't literally inject the module, since it determines the
current user in a different way (based on who's logged in rather
than on which Pod we're running), so this is currently a copy-paste
of most code in there, with a few import paths changed around.

Still, it at least shows where to copy changes from datasister
into, and vice versa: it maintains a similar overall structure, so
modifications made here can relatively easily be carried over to
datasister.

Note that this builds on #117, which wasn't merged yet. Thus, for reviewing it's probably easiest to just focus on these changes: https://github.com/solid/solid-panes/commit/760db5c52f41cb27827f8bdb0de7a86dbddea8ed


megoth:
This feature was made obsolete by the introduction of global dashboard by @timbl. Closing this PR for now.

Reopen if you want to continue this work.

PR #133, opened 2019-07-25
Add link to delete your account to preferences

Vinnl:
If necessary before tonight, this commit adds a "Delete account" link to the "Preferences".


megoth:
This will be added to the basicPreferences global dashboard pane instead.

PR #134, opened 2019-07-25
Dashboard as pane

megoth:
Introduces the dashboard w/global (context-free) panes that Tim introduced in https://github.com/solid/solid-panes/pull/123 as a pane, that is shown when you're authenticated on a Pod you control.

As a fallback (shown for guests - i.e. unauthenticated users and authenticated users who do not control the Pod they're visiting) we generate a homepage-like view on the fly (based on the [index.html that might be removed from Node-solid-server](https://github.com/solid/node-solid-server/pull/1276)).


dashboard/dashboardPane.ts
```diff
@@ -0,0 +1,69 @@
...
+
+function buildDashboard (container: HTMLElement, dom: HTMLDocument) {
+  const outliner = panes.getOutliner(dom)
+  outliner.showDashboard(container)
+}
+
+function buildHomePage (container: HTMLElement, subject: NamedNode) {
+  const wrapper = document.createElement('div')
+  container.appendChild(wrapper)
+  const shadow = wrapper.attachShadow({ mode: 'open' })
```
Vinnl, line 60: So the homepage is shown inside the data browser in the area reserved for the Pane, is that correct? I also just realised that Shadow DOM isn't supported in IE11, is that something we're OK with?

types.ts
```diff
@@ -54,3 +54,34 @@ interface NewPaneOptions {
   pane: PaneDefinition;
   refreshTarget: HTMLTableElement;
 }
+
+export interface SolidSession {
```
Vinnl, line 58: Time for another PR to DefinitelyTyped, I guess? :)

Vinnl (approved):

dashboard/dashboardPane.ts
megoth: The data browser is currently not working with IE at all - so in my mind, it is not something we can prioritize at the moment.

But we should be mindful of this in the future, so glad that you mentioned this.

megoth:
This was made obsolete by the changes merged into `master` from `dev` (https://github.com/solid/solid-panes/pull/141).

Reopen if necessary.

Issue #142, opened 2019-08-12
Should not list available panes within a global dashboard pane


megoth:
In the latest version of the data browser we've introduced global dashboards. These reuse parts of the outline manager that also reuses the listing of panes available. This list is probably just confusing in this version, and I propose that we remove this list from the global dashboard panes.

![AwesomeScreenshot-megoth-dev-inrupt-net--2019-08-12_1_54](https://user-images.githubusercontent.com/775139/62863308-b379c200-bd08-11e9-9a8f-4a486bc6f763.png)

Vinnl:
@timbl What do you think is the best way to detect whether the OutlineManager is rendering inside the dashboard?

Vinnl:
This issue was deprioritised since there was not a clear way to do this: removing the call to `renderPaneIconTray` when `options.pane.global` is `true` in `expandedHeaderTR` did remove that list, but it also caused the Pane not to get displayed at all. Since Tim did not consider this Issue that important, we're postponing it for now.

Of course, PRs still welcome if anybody else wants to pick this up :)

timbl:
Suggest setting the CSS visibility of the header TR to hidden, but leaving the structure of the DOM
 visibility: hidden;

timbl:
It isn't about the outline manager rendering inside the dashboard. It oi=is about the outline manager rendering a global pane, not a view pane.

megoth:
This was fixed in https://github.com/solid/solid-panes/pull/153

PR #153, opened 2019-08-15
Hide list of panes for global dashboard panes

megoth:
This should solve https://github.com/solid/solid-panes/issues/142

### DIFF
--- a/outline/manager.js
+++ b/outline/manager.js
@@ -381,7 +381,7 @@ module.exports = function (doc) {
   this.showDashboard = showDashboard
 
   function expandedHeaderTR (subject, requiredPane, options) {
-    function renderPaneIconTray (td) {
+    function renderPaneIconTray (td, hideList) {
       const paneShownStyle = 'width: 24px; border-radius: 0.5em; border-top: solid #222 1px; border-left: solid #222 0.1em; border-bottom: solid #eee 0.1em; border-right: solid #eee 0.1em; margin-left: 1em; padding: 3px; background-color:   #ffd;'
       const paneHiddenStyle = 'width: 24px; border-radius: 0.5em; margin-left: 1em; padding: 3px'
       const paneIconTray = td.appendChild(dom.createElement('nav'))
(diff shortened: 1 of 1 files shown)


outline/manager.js
```diff
@@ -381,7 +381,7 @@ module.exports = function (doc) {
   this.showDashboard = showDashboard
 
   function expandedHeaderTR (subject, requiredPane, options) {
-    function renderPaneIconTray (td) {
+    function renderPaneIconTray (td, hideList) {
```
Vinnl: To avoid making our API sprawl (`renderPaneIconTray(td, false, true, undefined, null, true)`) even worse, maybe make the second parameter an `options` object?

Vinnl (changes requested):
Ah, that works as well. Could you squash the commits? :)

outline/manager.js
megoth: I agree with the need to avoid API sprawl, but perhaps add an `options` object when we add more parameters? Feels a bit weird to do for one property.

Vinnl: But that means breaking the API, which is especially painful in a file that is not written in TypeScript :) (i.e. you'd not only have to update the parameters here, but also everwhere this function is called - though I guess that might be only in this file?)

megoth: Ah, right, that's a good argument. Ok, I'll make it into an option-parameter.

megoth:
@Vinnl I've pushed another commit that makes hideList into a property in the options parameter.

Please review again when you have time ^_^

Vinnl (approved):

megoth:
I think Tim had a couple of comments:

> Suggest setting the CSS visibility of the header TR to hidden, but leaving the structure of the DOM visibility: hidden;

https://github.com/solid/solid-panes/issues/142#issuecomment-521786138

> It isn't about the outline manager rendering inside the dashboard. It oi=is about the outline manager rendering a global pane, not a view pane.

https://github.com/solid/solid-panes/issues/142#issuecomment-521787197

I'll make another commit taking these comments into account, and then ask for another review.

megoth:
@timbl and I decided to take Tims feedback into https://github.com/solid/solid-panes/issues/154 and just merge this as-is.

PR #155, opened 2019-08-19
Allow loading .ttl files

Vinnl:
@timbl I've added the ability to import `.ttl` files into a Javascript file, which will automatically load the contents of the file into a Javascript variable. This is Webpack functionality, not official Javascript, so please double-check if that's desirable.

(Note that there also was a `versionInfo.js` file suddenly modified in this repo. Since I didn't know what that was, I did not include it in this commit. If it should be included, let me know @megoth.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8922,6 +8922,28 @@
         "safe-buffer": "^5.1.0"
       }
     },
+    "raw-loader": {
+      "version": "3.1.0",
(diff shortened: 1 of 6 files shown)


megoth:
I like it, makes the code easier to read.

Vinnl:
Given that we were already using the non-standardised `require`, I'm going to go ahead and merge this.

PR #157, opened 2019-08-22
A basic preferences global Pane that allows the user to indicate whether they're a Power User and/or a Developer

### LINKED ISSUE #144: Reintroduce basicPreferences
Tim introduced a new global dashboard pane called basicPreferences. This didn't quite work before our latest release, so we decided to remove it from the list of global dashboard panes and add it back later.

This issue is about adding it back, and fixing configurations that Tim added earlier (ask him about the specifics of this).

### LINKED ISSUE #145: Add developer and power user mode
This builds upon https://github.com/solid/solid-panes/issues/144.

We've talked about distinguishing between panes geared toward "normal" users, developers and power users. This issue is about adding the opt-in configurations of developer and power user mode (each being a checkbox) to the basicPreferences global dashboard pane.

To test these settings out, add `developer: true` and `powerUser: true` to some panes, and verify that they are not shown by default, but only when configured for a user.

I'll try to find the list of panes and which we decided should be which modes.

Vinnl:
This adds the code by Tim that uses [the Classifier](http://solid.github.io/solid-ui/Documentation/forms-intro.html) to set Power User or Developer mode.

Fixes #144, fixes #145.


dashboard/preferenceForm.ttl
```diff
@@ -0,0 +1,12 @@
+@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
+@prefix solid: <http://www.w3.org/ns/solid/terms#>.
+@prefix ui: <http://www.w3.org/ns/ui#>.
+@prefix : <#>.
+
+:this
+<http://purl.org/dc/elements/1.1/title> "Basic preferences" ;
+a ui:Form ;
+ui:part :categorizeUser;
```
timbl: The line `ui:part :categorizeUser;` is not needed  as that information is already in the following line.   Don't know why I originally wrote both.

timbl (changes requested):
Looks good apart from one small change

dashboard/preferenceForm.ttl
Vinnl: Added a commit that fixes that.

dashboard/basicPreferences.source.ts
```diff
@@ -1,17 +1,22 @@
 import { PaneDefinition } from '../types'
 import UI from 'solid-ui'
-import $rdf, { NamedNode } from 'rdflib'
+import * as $rdf from 'rdflib'
```
megoth: Any specific reason for why we're doing this instead of just `import { NamedNode } from 'rdflib'`? We only need that for this file.

dashboard/basicPreferences.source.ts
```diff
@@ -1,17 +1,22 @@
...
 
 const kb = UI.store
 
 export const basicPreferencesPane: PaneDefinition = {
   icon: UI.icons.iconBase + 'noun_Sliders_341315_000000.svg',
   name: 'basicPreferences',
-  label: () => null,
+  label: (subject) => {
```
megoth, line 10: We want to expose this as a normal pane? I thought we only wanted to expose it through the global dashboard?

dashboard/basicPreferences.source.ts
```diff
@@ -21,50 +26,88 @@ export const basicPreferencesPane: PaneDefinition = {
...
+      }
     }
+    const preferencesForm = kb.sym('urn:uuid:93774ba1-d3b6-41f2-85b6-4ae27ffd2597#this')
+    loadData(preferencesForm, preferencesFormText)
+
+    const ontologyExtra = kb.sym('urn:uuid:93774ba1-d3b6-41f2-85b6-4ae27ffd2597-ONT')
+    loadData(ontologyExtra, ontologyData)
+
+    async function doRender () {
+      var context = await UI.authn.logInLoadPreferences({ dom, div: container })
```
megoth: Why using `var` instead of a more specific syntax such as `const`?

dashboard/ontology-extra.ttl
```diff
@@ -0,0 +1,33 @@
...
+@prefix ui: <http://www.w3.org/ns/ui#>.
+@prefix vcard: <http://www.w3.org/2006/vcard/ns#>.
+@prefix : <#>.
+
+solid:User a rdfs:Class;
+  rdfs:label "user"@en, "utilisateur"@fr;
+  rdfs:comment """Any person who might use a Solid-based system""";
+  rdfs:subClassOf foaf:Person, schema:Person, vcard:Indivudual.
+
+solid:NewUser a rdfs:Class;
```
megoth: I think we should use these files instead of the inline versions in basicPreferences.source.ts.

Now we have inconsistent models, as solid:NewUser is commented out in basicPreferences.source.ts.

megoth (changes requested):
Just some questions and smaller things to fix - looks good ^_^

dashboard/ontology-extra.ttl
Vinnl: Yes, but the functionality to do that wasn't merged yet in this branch, and I didn't want to rebase again because that caused Tim to locally delete and re-clone this repo locally the previous time. I'll delete the `.ttl` files for now, since they don't do anything currently, and we can recreate them later.

dashboard/basicPreferences.source.ts
Vinnl: Tim uses `var` usually, and I hadn't noticed. I'll review PRs better before submitting them next time - fixed now.

dashboard/basicPreferences.source.ts
Vinnl: Ah, sorry, I wasn't aware that this wasn't necessary for global Panes - I thought it was commented out because it wasn't ready yet. Fixed now.

dashboard/basicPreferences.source.ts
Vinnl: `$rdf.parse` is called as well, but we can still import both, of course.

Vinnl:
Thanks Arne, I've [incorporated your suggestions](https://github.com/solid/solid-panes/pull/157/files/999d7bc59536d718518a5cf86aaa8ab69766eafd..d2a2ff2a4140a46113fdf14739a2434ad2a446f1).

megoth (approved):
LGTM ^_^

megoth:
@Vinnl Could you fix the conflicting file, then merge? You've fixed Tim's feedback, so I reckon it looks good for him as well ^_^

PR #163, opened 2019-08-27
Add account deletion link(s) to preferences

### LINKED ISSUE #146: Add link to delete user profile from basicPreferences
This builds upon https://github.com/solid/solid-panes/issues/144.

This should add the link that allows users to delete their accounts and Pods on NSS.

(This used to be shown for authenticated users on the generated index.html that we had for NSS before.)

Vinnl:
Note that this method assumes that the server is running Node Solid
Server, or at least provides the deletion link at
`/account/delete`.

It makes `HEAD` requests to what would be the deletion pages in single-user and multi-user Node Solid Server setups, respectively, as suggested at https://github.com/solid/solid-panes/issues/146#issuecomment-524858414.

Fixes #146.

### DIFF
--- a/dashboard/basicPreferences.source.ts
+++ b/dashboard/basicPreferences.source.ts
@@ -1,12 +1,14 @@
 import { PaneDefinition } from '../types'
 import UI from 'solid-ui'
-import { NamedNode, parse } from 'rdflib'
+import namespace from 'solid-namespace'
+import { NamedNode, parse, namedNode, IndexedFormula } from 'rdflib'
 import { renderTrustedApplicationsOptions } from './trustedApplications/trustedApplicationsPane'
 
 import preferencesFormText from './preferencesFormText.ttl'
 import ontologyData from './ontologyData.ttl'
 
 const kb = UI.store
+const ns = namespace({ namedNode } as any)
 
 export const basicPreferencesPane: PaneDefinition = {
   icon: UI.icons.iconBase + 'noun_Sliders_341315_000000.svg',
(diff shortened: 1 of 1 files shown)


dashboard/basicPreferences.source.ts
```diff
@@ -106,3 +109,54 @@ solid:PowerUser a rdfs:Class;
...
+ *
+ * This function infers the deletion link by assuming the URL structure of Node Solid server.
+ * In the future, Solid will hopefully provide a standardised way of discovering the deletion link:
+ * https://github.com/solid/data-interoperability-panel/issues/18
+ *
+ * @param server Pod server containing the user's account.
+ * @returns URL of the page that Node Solid Server would offer to delete the account, or null if
+ *          the URLs we tried give invalid responses.
+ */
+async function getDeletionUrlForServer (server: string): Promise<string | null> {
```
megoth, line 103: An alternative version could be to simple give a list of URLs to check and use `Promise.allSettled`, instead of doing one by one - but this is a hack in any case, and I don't have a problem with this solution (which could of course save one request).

megoth (approved):
LGTM

megoth:
Please fix conflicting files, then merge.

dashboard/basicPreferences.source.ts
Vinnl: The disadvantage of that is that it will make all HTTP requests at the same time, whereas with the current code it will not make the second request if the first one resolves. (Though it'd probably be good to first try it for the multi-user setup, which is more likely - I'll change that.)

(Edit: ah, I see your final sentence says the same, sorry :P)

An array of URLs could work, but that's optimising for being able to add more URLs later - I'd hope that that would not be necessary, thanks to https://github.com/solid/data-interoperability-panel/issues/18.

PR #165, opened 2019-08-27
Move trustedApplications into basicPreferences

### LINKED ISSUE #151: Move TrustedApplications to the basicPreferences Pane
As requested by Tim.

Depends on #144.

Vinnl:
This removes everything but the `render` function from the trustedApplications Pane definition, and exports that. `trustedApplicationsPane.ts` looks like a large diff, but the only change is the removal of the unused `subject` parameter, and changing the indentation and function name.

Fixes #151.


megoth (approved):
LGTM

The only problem I saw was some missing styles on headers - but this is not a problem with the work here, this is the result of earlier work. I've created an issue for it - https://github.com/solid/mashlib/issues/66.

PR #166, opened 2019-08-27
Separating dashboard and outline

### LINKED ISSUE #154: Global panes should replace the view for normal panes
Right now we're doing a hack that replaces the view of the pane with the global panes, but the list of available panes for the current resource is still available.

We should make it properly replace the view of the pane, and have an x-icon or something that makes it easy to close the global panes to get back to the "normal" panes for the current resource.

megoth:
No reason to reuse elements from the outline container that is used for displaying panes for a given subject.

This commit separates the logic for the two, and handles switching between them.

Not done any styling on cancel button, as I want to know how Tim wants to have it first.

This fixes https://github.com/solid/solid-panes/issues/154

This has some breaking changes in API that [mashlib](https://github.com/solid/mashlib/) uses, and mashlib will require these commit to continue functioning.

### DIFF
--- a/outline/manager.js
+++ b/outline/manager.js
@@ -42,7 +42,7 @@ module.exports = function (doc) {
   var outlineElement = this.outlineElement
 
   this.init = function () {
-    var table = dom.getElementById('outline')
+    var table = getOutlineContainer()
     table.outline = this
   }
 
@@ -272,9 +272,11 @@ module.exports = function (doc) {
   } // outlinePredicateTD
 
 /** Render Tabbed set of home app panes
+ * @param {Object} [options] A set of options you can provide
+ * @param {string} [options.selectedTab] To open a specific dashboard pane
  * @returns Promise<{Element}> - the div
 */
-  async function globalAppTabs (selectedTab) {
+  async function globalAppTabs (options = {}) {
     console.log('globalAppTabs @@')
     const div = dom.createElement('div')
     const me = UI.authn.currentUser()
@@ -297,19 +299,20 @@ module.exports = function (doc) {
       thisOutline.GotoSubject(item.subject || me, true, pane, false, undefined, table)
     }
 
-    const options = {dom,
+    div.appendChild(UI.tabs.tabWidget({
+      dom,
       subject: me,
       items,
       renderMain,
       renderTab,
       ordered: true,
       orientation: 0,
-      backgroundColor: '#eeeeee',
-      selectedTab} // black?
-    // options.renderTabSettings = renderTabSettings  No tab-specific settings
-    div.appendChild(UI.tabs.tabWidget(options))
+      backgroundColor: '#eeeeee',  // black?
+      selectedTab: options.selectedTab
+    }))
     return div
   }
+  this.getDashboard = globalAppTabs
 
   async function getDashboardItems () {
     const me = UI.authn.currentUser()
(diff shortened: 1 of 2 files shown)


megoth:
The corresponding changes in mashlib can be found at https://github.com/solid/mashlib/pull/67

outline/manager.js
```diff
@@ -365,19 +368,73 @@ module.exports = function (doc) {
...
-  async function showDashboard (container, unselectCurrentPane, globalPaneToSelect) {
-    container.innerHTML = ''
-    // console.log(container)
-    const currentPane = dom.querySelector('#outline .paneShown')
-    if (unselectCurrentPane && currentPane) {
-      // eslint-disable-next-line no-undef
-      // currentPane.dispatchEvent(new Event('clglobalAppTabsick'))
+  /**
+   *
+   * @param {HTMLElement} container The element that the dashboard should append dashboard to
```
Vinnl: This param seems to have been replaced by `getDashboardContainer`?

```suggestion
```

outline/manager.js
```diff
@@ -365,19 +368,73 @@ module.exports = function (doc) {
...
+   */
+  async function showDashboard (options = {}) {
+    const dashboardContainer = getDashboardContainer()
+    const outlineContainer = getOutlineContainer()
+    // reuse dashboard if already children already is inserted
+    if (dashboardContainer.childNodes.length > 0 && options.pane) {
+      outlineContainer.style.display = 'none'
+      dashboardContainer.style.display = 'inherit'
+      const tab = dashboardContainer.querySelector(`[data-name="${options.pane}"]`) ||
+        dashboardContainer.querySelector('[data-name]')
```
Vinnl: I presume `data-name` is already used somewhere? If not, it's a bit odd to have a data attribute with the same name as a regular attribute (i.e. `name`), and something like `data-global-pane-name` or something would be clear. But no biggy if it's already used.

Vinnl (approved):
Overal looks good; fine with merging if the JSDoc is fixed.

outline/manager.js
megoth: You're correct, I should use a more specific name than one that resembles standard attribute `name`. Will use `data-global-pane-name` instead.

outline/manager.js
```diff
@@ -365,19 +372,80 @@ module.exports = function (doc) {
...
+      outlineContainer.style.display = 'none'
+      dashboardContainer.style.display = 'inherit'
+      const tab = dashboardContainer.querySelector(`[data-global-pane-name="${options.pane}"]`) ||
+        dashboardContainer.querySelector('[data-global-pane-name]')
+      if (tab) {
+        tab.click()
+        return
+      }
+      console.warn('Found no tabs in global dashboard to open')
+      dashboardContainer.innerHTML = ''
```
timbl: If no specific tab is specified by the caller, why not let the tabs widget just select the default tab?  Doesn't `dashboardContainer.innerHTML = ''` zap the dashboard?

timbl (approved):

outline/manager.js
megoth: It was to catch the case where a user tries to open a dashboard pane that isn't available - it should not happen, so was just to fetch that error.

But I've updated with a commit that tries to make this clearer. And yes, the fallback is that tabs will open the first one, so removed some redundant logic in mashlib code.

Issue #167, opened 2019-08-28
Only show panes based on the type of user


megoth:
With https://github.com/solid/solid-panes/issues/145 we introduced the ability for users to set themselves as developer and/or power user. The data browser should reflect this setting by limiting the panes you are presented based on which user type you've set yourself as.

megoth:
Will add properties to panes based on [the overview we've worked on](https://team.inrupt.net/public/mashlib//Panes%20(developer%20and%20power%20user%20mode).pdf).

megoth:
This was fixed in the following PRs:
- https://github.com/solid/solid-ui/pull/112
- https://github.com/solid/folder-pane/pull/3
- https://github.com/solid/issue-pane/pull/1
- https://github.com/solid/meeting-pane/pull/1
- https://github.com/solid/solid-panes/pull/168

PR #168, opened 2019-08-29
Adding audience to panes and filtering based on that and user roles

megoth:
This is dependent on solid/solid-ui#112, so do not merge this until it has been merged and released and the PR has been updated with corresponding release version.


megoth:
This PR should also include an update to issue-pane after https://github.com/solid/issue-pane/pull/1 has been merged.

megoth:
This PR should also include an update to meeting-pane after https://github.com/solid/meeting-pane/pull/1 has been merged.

megoth:
This PR should also include an update to folder-pane after https://github.com/solid/folder-pane/pull/3 has been merged.

megoth:
This is part of fixing https://github.com/solid/solid-panes/issues/167

timbl (changes requested):
It is important that when a user is looking at an object (like a meting or a bank transaction ) where there is a hand-coded view, then they must be given that view, whether or not they are an expert user. It is very bad to deprive them of the proper view and give them a raw data view!   We can I suppose detect that as the highest priority pane in the way the pane system works at the moment,  though we cloudl cahnge that to haveing q values for each pane.

defaultPane.js
```diff
@@ -5,14 +5,17 @@
...
+const UI = require('solid-ui')
+const panes = require('pane-registry')
+const ns = UI.ns
 
 module.exports = {
   icon: UI.icons.originalIconBase + 'about.png',
 
   name: 'default',
 
+  audience: [ns.solid('Developer')],
```
Vinnl, line 17: I'm not quite sure what this Pane does, but given that it's named "Default": is this the one that will be visible if no other Pane matches? In other words: will the user get an empty screen if this one's not present either? If so, maybe it should just do what it does today if the user's a Developer, and display a message about why the user doesn't see anything if not.

outline/manager.js
```diff
@@ -490,8 +475,7 @@ module.exports = function (doc) {
                 ico.style = paneHiddenStyle
               }
 
-                // If the view already exists, remove it
-              state = 'paneShown'
+              // If the view already exists, remove it
```
Vinnl: It's not quite clear to me what line this comment applies to, but if it was the line `state = 'paneShown'`, then I suppose this comment should be removed too :)

Vinnl (changes requested):
LGTM; setting it to "Request changes" to make sure the upgrade to a new major version of solid-ui with https://github.com/solid/solid-ui/pull/112 is included.

defaultPane.js
megoth: I think the name is just misleading - I suspect it was one of the first panes added, and the idea was to give you a little bit of info about the resource. But this info is still designed for developers.

If no panes are found, it will show `internalPane` instead. (This is also noted as developer-oriented, but this is hard-coded to be shown if no other panes are available.)

megoth:
@Vinnl and @timbl can you accept the reviews, so that I can do the appropriate merges?

Vinnl (approved):

outline/manager.js
```diff
@@ -380,8 +380,21 @@ module.exports = function (doc) {
 
   async function getRelevantPanes (panes, subject, dom) {
     const relevantPanes = panes.list.filter(pane => pane.label(subject, dom) && !pane.global)
+    if (relevantPanes.length === 0) {
+      // there are no relevant panes, simply return default pane (which ironically is internalPane)
```
timbl, line 384: ¿¿¿ eh??? The `default` pane and the `internals` pane should be quite different.  The default pane should the normal properties in a list, excluding things deemed to be internal like the URI itself, the record of attempts to load the URI etc.  Those things are shown in the internals pane, which also allows you to refresh the local triple store, delete the file, etc.  The default pane is not the internal pane.

timbl (approved):

outline/manager.js
megoth: The irony is in the naming - defaultPane sounds like it should be the _default_ pane that is shown.

Issue #170, opened 2019-09-04
Can edit Trusted App URL but does not save


brownhoward:
When viewing the list of Trusted Apps in the Data Browser, I am allowed to edit the Trusted App URLs. However, when I click 'Update', the edited URL is not saved. Rather a new Trusted App record is created.

Either:
1. The edited URL should be saved correctly; or,
2. The URL should be made read-only.

I suggest we go with 2. as it is easy enough to delete and add a Trusted App, and we would need to somehow confirm the URL is correct.

PR #177, opened 2019-09-18
Updates same row when entering new domain for trusted app

megoth:
Fix for https://github.com/solid/solid-panes/issues/170

Also contains some clean-up of syntax (scout rule)

### DIFF
--- a/dashboard/trustedApplications/trustedApplicationsPane.ts
+++ b/dashboard/trustedApplications/trustedApplicationsPane.ts
@@ -164,7 +164,6 @@ function createApplicationEntry (
       .filter(checkbox => checkbox.checked)
       .map(checkbox => checkbox.value)
 
-    console.log(trustedApplicationState.origin, trustedApplicationState.formElements.origin)
     const deletions = getStatementsToDelete(trustedApplicationState.origin || origin, subject, kb, ns)
     const additions = getStatementsToAdd(origin, generateRandomString(), modes, subject, ns);
     (kb as any).updater.update(deletions, additions, handleUpdateResponse)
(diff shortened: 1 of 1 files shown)


dashboard/trustedApplications/trustedApplicationsPane.ts
```diff
@@ -109,65 +109,78 @@ function createApplicationEntry (
...
     }
 
-    var modes = trustedApplicationState.formElements.modes
+    const modes = trustedApplicationState.formElements.modes
       .filter(checkbox => checkbox.checked)
       .map(checkbox => checkbox.value)
 
-    var deletions = getStatementsToDelete(origin, subject, kb, ns)
-    var additions = getStatementsToAdd(origin, generateRandomString(), modes, subject, ns);
+    console.log(trustedApplicationState.origin, trustedApplicationState.formElements.origin)
```
Vinnl:
```suggestion
```

Forgot to remove this :)

Vinnl (approved):
I'm not quite sure what `createContainer`, `createElement` and `createText` do, but I'm assuming it's pretty much what I'd expect. Still not entirely clear to me what exactly this change does, but I'm trusting it'll be the right thing.

One request for next time though: if you do a general code cleanup, could you do that in a separate commit? Then I can review the relevant change by itself, and just scan over the cleanup diffs. Thanks!

megoth:
> One request for next time though: if you do a general code cleanup, could you do that in a separate commit? Then I can review the relevant change by itself, and just scan over the cleanup diffs. Thanks!

Ah, yes, that was lazy of me. Will do next time ^_^

PR #186, opened 2019-10-09
Reintroducing internalPane to all users

megoth:
People can't delete files without this pane for now. We'll add another view for this later on, but add this as a quick fix for now.


megoth:
I'll close this since it's so old.

PR #187, opened 2019-10-09
Introducing prettier-standard

megoth:
Reusing existing eslint-rules and enforcing them consistently with pre-commit check.

Removed tslint.json as prettier-standard should handle our needs

There are no big changes except the ones mandated by prettier-standard, so the only files that should be needed to check are:
- .eslintignore
- package-lock.json
- package.json


Vinnl (approved):
Cool, I didn't know Prettier did Markdown too.

I'm not sure about the removal of the linter though, as they do not only enforce code conventions, but can also prevent bugs - e.g. it can warn you of unhandled Promises.

That said, we'd probably want to switch to ESLint anyway, so as far as I'm concerned I'm fine with re-introducing it later, and just merge this.

PR #195, opened 2019-11-20
Moving src files to src folder

megoth:
Is a pattern used in other big projects, and makes distinguishing source files from configuration and documentation files much easier.


package.json
```diff
@@ -13,7 +13,7 @@
     "lint": "prettier-standard --lint '**/*'",
     "test-watch": "npm run lint && jest --onlyChanged --watch",
     "test": "npm run lint && jest",
-    "pre": "./timestamp.sh > versionInfo.js"
+    "pre": "./timestamp.sh > src/versionInfo.js"
```
Vinnl, line 16: I'd leave this one in the top-level, since it's auto-generated rather than a source file we author.

Vinnl (approved):
GOD YES

package.json
megoth: It's referred to in src/index.js, so I would prefer to have it in src folder - I at least don't want to have it on root, but could have another folder to put it in since it's a bit different than the source files in src

PR #196, opened 2019-11-20
Changing to transpiling source files with Babel

megoth:
But keeping Webpack for those who want to still bundle files


package.json
```diff
@@ -2,14 +2,16 @@
   "name": "solid-panes",
   "version": "2.0.4",
   "description": "Solid-compatible Panes: applets and views for the mashlib and databrowser",
-  "main": "dist/main.js",
+  "main": "lib/index.js",
   "scripts": {
-    "build": "npm run clean && npm run build-dist && npm run build-dev",
+    "build": "npm run clean && npm run build-dist && npm run build-dev && npm run build-lib && npm run build-types",
```
Vinnl, line 7: Doesn't `npm run build-dist` already build the Webpack types? And if not, wouldn't it be better for it to do that?

Vinnl (approved):

package.json
Vinnl: Never mind - you probably don't need typings if you're including the bundle directly with a `<script>` tag.

PR #197, opened 2019-11-20
Generate typing files

megoth:
This is far from complete, but starts the work of eventually having types for all of solid-panes (and other related projects)


src/versionInfo.ts
```diff
@@ -0,0 +1,24 @@
+export default {
```
Vinnl, line 1: Does the script that generates this file know that it should generate a `.ts` file in the future?

Vinnl (approved):
Looks fine on a quick scan, but make sure to test this properly when including in a different project. In one of our many repositories (can't remember which one), enabling TypeScript led to only an `.index.d.ts` file being included, which referred to other files that did _not_ have type definitions, thus failing the build.

src/versionInfo.ts
megoth: Yes, you can see the script `pre` in package.json

Vinnl: Ah, oops, missed that - thanks!

PR #199, opened 2019-11-21
Changing from dom to context

megoth:
Trying to keep the changes to a minimum, but this is a breaking change.

Major changes in methods:

label (subject) -> label (subject, context)
mintNew (options) -> mintNew (context, options)
render (subject, dom, options) -> render (subject, context, options)

Major changes in variables:

dom -> context.dom
UI.store -> context.session.store
panes -> context.session.paneRegistry

Some refactoring to clean up TypeScript files and allowing them to be compiled.

Some casting necessary, but added @@ TODO on all of them.


src/defaultPane.js
```diff
@@ -21,9 +20,15 @@ module.exports = {
...
+  render: function (subject, context) {
+    var dom = context.dom
+
     var filter = function (pred, inverse) {
-      if (typeof panes.internal.predicates[pred.uri] !== 'undefined') {
+      if (
+        typeof context.session.paneRegistry.byName('internal').predicates[
+          pred.uri
+        ] !== 'undefined'
+      ) {
```
Vinnl, line 31: Is this the result of `eslint --fix`? I don't want to get into some bikeshedding thing, but this does look... Weird? Or is it just me?

Vinnl, line 31: Is this the result of `eslint --fix`? I don't want to be all bikesheddy and all, but this looks... Weird? Or is it just me?

src/imagePane.js
```diff
@@ -45,7 +45,8 @@ module.exports = {
     return 'view'
   },
 
-  render: function (subject, myDocument) {
+  render: function (subject, context) {
+    const myDocument = context.dom
```
Vinnl, line 49: `dom` isn't actually the `HTMLDocument` object, right? i.e. `myDomElement` would be more descriptive of what it actually is?

Vinnl (approved):
Looks good, only comments I have to offer are some bikeshedding disguised as not-bikeshedding. Hopefully this will be useful to Thom, and Joep's work will certainly be useful in being able to remove many of the `as any`'s here.

src/imagePane.js
megoth: `dom` is the actual `HTMLDocument`, yes.

src/defaultPane.js
megoth: Hmm, don't think this should be necessary. Could be because of prettier and the extra length of the line. I'll revert the uses of newlines here for now.

src/imagePane.js
Vinnl: Huh, if that's the case, then it's a bit weird to pass it around in the context, right, given that it's available globally anyway?

src/defaultPane.js
megoth: Nvm, Prettier won't let me. That's the final nail in the coffin then, will revert it in a later PR ^_^

src/imagePane.js
megoth: Not necessarily - in Electron you have multiple documents, and being able to specify which DOM that should be used is useful.

Vinnl: Ah, TIL!

PR #201, opened 2019-11-25
A set of scripts proposed to become the default in mashlib projects

megoth:
Before git commit: Checking lint
Before git push: Running tests
Before npm publish: Running tests and building files
After npm publish: Pushing tags

Also ignoring the built versionInfo file so that it does not require a separate git commit when we've built our files for publishing.


Vinnl (approved):

timbl (approved):

timbl:
lgtm

PR #202, opened 2019-11-27
Adding support for default in ES Modules

megoth:
When adding views it will check for the default property.

This removes the need for `.source.ts`-postfix pattern that we've used when migrating TS-files earlier.


Vinnl (approved):
This is vastly preferable to me over the previous workaround. :+1:

timbl:
So we are changing the API for a all panes to a default export?

Vinnl:
@timbl It's an additional API, so panes that use a CommonJS export continue to work, but panes that use regular Javascript exports (and use the default export) would work after this PR as well.

src/internal/internalPane.ts
```diff
@@ -1 +1,255 @@
...
+          noun,
+          function () {
+            if (
+              !confirm(
+                `Are you sure you want to delete ${subject}? This cannot be undone.`
+              )
+            ) {
+              return
+            }
+            // @@ TODO Remove casing of store.fetcher
```
timbl, line 113: To remove  a casting of store.fetcher and store.updater, how about making superclasses of Store, like ConnectedStore or OnlineStore for the one which has a fetcher, and LiveStore superclass of OnlineStore which would also have an updater ?

timbl: Or is it too valuable to be able to make store and attach its helper objects later?

timbl (approved):

src/internal/internalPane.ts
megoth: Was thinking the same thing, and have a local commit that's about to be a PR for that. Can push it tomorrow.

megoth: Have pushed a PR to `pane-registry` with proposed changes - https://github.com/solid/pane-registry/pull/5

PR #203, opened 2019-11-27
Migrating tabbed view as example of how migration could look like

megoth:
Want to show Tim how I'm thinking this can be done in general


src/tabbed/tabbedPane.ts
```diff
@@ -0,0 +1,66 @@
...
+        renderMain: prepareRenderMainFn(context),
+        renderTab: prepareRenderTabFn(subject, context),
+        backgroundColor: store.anyValue(subject, ns.ui('backgroundColor')) || '#ddddcc'
+      }))
+    })()
+    return div
+  }
+}
+
+function prepareRenderTabFn (subject: NamedNode, context: DataBrowserContext): (div, item) => void {
```
timbl: As a question of code style, simpler is better.   A one-line `renderTab` in line where it has its state used is better than a four-line one far from where it is used.    A function is simpler than a function generator function.

timbl (changes requested):
The changes to ts seem fine in general.  Suggest that render() be simplified back to simpler functions as it had been.  (Are we sure we want to use export default?)

megoth:
@timbl Have implemented your requested changes, hopefully it looks good now ^_^

src/tabbed/tabbedPane.ts
```diff
@@ -0,0 +1,66 @@
...
+
+function prepareRenderMainFn (context: DataBrowserContext): (containerDiv, item) => void {
+  return (containerDiv, item) => {
+    containerDiv.innerHTML = ''
+    const table = containerDiv.appendChild(context.dom.createElement('table'))
+    ;(context.getOutliner(context.dom) as any).GotoSubject(item, true, null, false, undefined, table)
+  }
+}
+
+export default TabbedPane
```
timbl, line 58: Are you sure that export default is what we need -- won't that give is a .default. in the path when we use a require()  ?

timbl (changes requested):

src/tabbed/tabbedPane.ts
megoth: I forgot to note that this PR is dependent on the changes in https://github.com/solid/solid-panes/pull/202#discussion_r359061029 - with it we will support both CJS and ES modules.

megoth:
@timbl Am I to understand that you approve the rest of this PR now that https://github.com/solid/solid-panes/pull/202 is merged?

PR #205, opened 2019-11-28
Found a bug in attachPane

megoth:
The bug ended up sending an empty string to kb.sym, which is not allowed


timbl (changes requested):
This code surely
- needlessly changes a map (functional) to a case statement (procedural) 
- returns strings instead of RDF objects

Suggest  
```
var uri = { ...map... } [subject.uri] 
return uri ? kb.sym(uri) :  kb.any(subject, UI.ns.ui('sortBy'))
````
or something like that

PR #207, opened 2019-12-18
Migrated SharingView to TypeScript

megoth:
A preliminary PR for the work on https://github.com/solid/solid-ui/issues/121, so the fix with actual changes don't have confusing parts as part of migrating the file to TypeScript.


megoth:
Tim and I discussed this matter, and for these kinds of PRs (simple migrations with no changed behavior) he does not need to review them (we can simply merge them).

PR #211, opened 2020-01-09
Re-adding trusted applications as view

megoth:
Is needed to be able to render trustedApplications in Solid UI (part of solid/solid-ui#121)


src/profile/profilePane.utils.ts
```diff
@@ -0,0 +1,10 @@
+import { IndexedFormula, Node } from 'rdflib'
+import { Namespaces } from 'solid-namespace'
+
+export function getLabel (subject: Node, kb: IndexedFormula, ns: Namespaces) {
+  var types = kb.findTypeURIs(subject)
+  if (types[ns.foaf('Person').uri] || types[ns.vcard('Individual').uri]) {
+    return 'Your Profile'
+  }
```
timbl: I don't understand the logic here.  If this is a view, then the label will  be non-null if the view applies to the subject.  This condition means that the subject is a person.  The view will be of that person't profile, not the user's profile. 

If it is now a view but a applet, -- view/edit my profile -- then the subject is irrelevant. It should alway s be available, no matter what the user is looking at.

timbl (changes requested):

michielbdejong:
Can you give some more context?
- Why are we readding trusted applications? I didn't know they had been removed?
- What is the relation with the green pluses in the Sharing pane?
- How does this relate to the removal of trusted applications in [the december version of the Solid spec](https://github.com/solid/specification/issues/135#issuecomment-566418253)?

src/profile/editProfile.view.ts
```diff
@@ -0,0 +1,147 @@
...
+          // @@ TODO Remove casting of store.updater
+        } else if ((store.updater as any).editable(profile.uri, store)) {
+          editableProfile = profile
+        } else {
+          statusArea.appendChild(widgets.errorMessageBlock(dom, `⚠️ Your profile ${profile} is not editable, so we cannot do much here.`, 'straw'))
+          return
+        }
+
+        comment(`Everything you put here will be public.
+        There will be other places to record private things..`)
```
TallTed:
```suggestion
        There will be other places to record private things.`)
```

src/profile/editProfile.view.ts
```diff
@@ -0,0 +1,147 @@
...
+
+        heading('Your contact information')
+
+        main.appendChild(paneDiv(context, me, 'contact'))
+
+        heading('People you know who have webids')
```
TallTed:
```suggestion
        heading('People you know who have WebIDs')
```

src/profile/editProfile.view.ts
```diff
@@ -0,0 +1,147 @@
...
+
+        comment(`This is your public social network.
+        Just put people here you are happy to be connected with publicly
```
TallTed:
```suggestion
        Only put people here to whom you are happy to be publicly connected.
```

src/profile/profile.view.ts
```diff
@@ -1,46 +1,30 @@
...
- ** worth the subjectjoing solid for
- ** - informative
- **
- ** Usage: paneRegistry.register('profile/profilePane')
- ** or standalone script adding onto existing mashlib.
+/**
+ * Display A Public Profile Pane
+ *
+ * This is the subject's primary representation in the world.
+ * When anyone scans the QR code of their webid on theor card, it takes gthem
```
TallTed:
```suggestion
 * When anyone scans the QR code of their WebID on their card, it takes them
```

src/profile/profile.view.ts
```diff
@@ -1,46 +1,30 @@
...
+ * to here and here alone.  Thiks has better be good.  This has better be
```
TallTed:
```suggestion
 * to here and here alone.  This has better be good.  This has better be
```

src/profile/profile.view.ts
```diff
@@ -1,46 +1,30 @@
...
+ * worth the subjectjoing solid for
```
TallTed:
```suggestion
 * worth the subject joining solid for
```

src/profile/profilePane.utils.ts
megoth: This logic is nevertheless outdated, as it was only used in editProfile, and we don't need to label it as it is only used by the dashboard anyway.

I can't recall the exact reasoning for the logic in the code above, but perhaps @Vinnl does? (I'll ask him to comment on Monday.)

megoth:
> Can you give some more context?
> 
>     * Why are we readding trusted applications? I didn't know they had been removed?

It was not, but it had been rewritten into a function instead of a view earlier - this PR is about readding trustedApplications _as a view_. I've updated the name of the PR to reflect this.

> 
>     * What is the relation with the green pluses in the Sharing pane?

When you want to add an application in the sharing view, we also render the trustedApplications view, so that you can manage trusted applications from that view. (I'm not sure if this is how we want to do it tbh, but Tim added it like this earlier and I didn't want to change it as part of this PR.)

> 
>     * How does this relate to the removal of trusted applications in [the december version of the Solid spec](https://github.com/solid/specification/issues/135#issuecomment-566418253)?

IMO we need to remove Trusted applications view from the data browser in the long run, but we'll leave it as is since NSS still supports it for now. At that point, we need to fetch the list of applications users can add from somewhere else.

src/profile/profilePane.utils.ts
Vinnl: I don't know either, [it was added by Tim two years ago](https://github.com/solid/solid-panes/commit/e3d160b31ee4044afd9a080a2f896a1fb516abfa#diff-c6baf3719ead006cc587c2d4d7368fa6R22-R25).

src/profile/profile.view.ts
```diff
@@ -1,46 +1,30 @@
...
- ** - informative
- **
- ** Usage: paneRegistry.register('profile/profilePane')
- ** or standalone script adding onto existing mashlib.
+/**
+ * Display A Public Profile Pane
+ *
+ * This is the subject's primary representation in the world.
+ * When anyone scans the QR code of their WebID on their card, it takes them
+ * to here and here alone.  This has better be good.  This has better be
```
TallTed:
```suggestion
 * to here and here alone.  This had better be good.  This had better be
```
Sorry I missed these before!

src/profile/profilePane.utils.ts
timbl: That was logic for a view (pane), not an app.    Trusted applications is an app.  Hence the nonsense. And the label then was "friends" -- which could apply to anyone when looking at someone else's profile, not "your profile", which does not make sense when looking at someone else.   But now I see more or less the route by which we got here.

timbl: One reason why we find it difficult to look at this is that it has been split off as too small a module.  The entire code of the module is 
```
export function getLabel (subject: Node, kb: IndexedFormula, ns: Namespaces) {
  var types = kb.findTypeURIs(subject)
  if (types[ns.foaf('Person').uri] || types[ns.vcard('Individual').uri]) {
    return 'Your Profile'
  }
  return 'Edit your profile' // At the moment, just allow on any object. Like home pane
}
```
The name of the function, getLabel doesn't describe what it does, which is something like "return a pane label iff the subject is a person".   Which is what we do not want.  When you split off  functionality into  another module, it is really important that the reader of the calling module  understands just what is going on in the call.   This is in fact a logical condition for the app to appear, and at the top level of the app is a good place for it.  Its a top level thing.  Plus, there is a cognitive load and a quantum of effort required when the reader switches to the called module to see  what is  going on.  Don't waste that effort for two lines, keep it for 20 or more lines.   I suggest
- reintegrate those two lines into the calling code
- remove the test on the subject.'s class as it is irrelevant to an app pane
- just return a constant string as a label 'profile' say.    (module I18n later)

timbl (changes requested):

megoth:
Tim and I have discussed this offline, and he's ok with the latest changes that removes the getLabel function altogether.

Issue #213, opened 2020-02-03
Outliner crashes trying to search for embedded graph in justification code


timbl:
Strange bug -- justification code is not used now and simplest thing to remove it.
The code used a new (one statement) graph as a subject in the triple store to store reasons for believing things.

timbl:
![image_2020_02_03T18_06_06_323Z](https://user-images.githubusercontent.com/1254848/73681346-28143480-468c-11ea-9fcd-3df96ad71cb4.png)

melvincarvalho:
Curious indeed.  Possibly a similar issue

> It seems to happen when trying to parse any RDF resource. Needless to say, this makes the data browser unusable.

https://github.com/linkeddata/rdflib.js/issues/377

melvincarvalho:
https://github.com/linkeddata/rdflib.js/blob/d5000f58258ceb8bf3352be24e919b313cf66abd/src/factories/canonical-data-factory.ts#L178

```JavaScript
  /** Stringify a {term} to n-quads serialization. */
  termToNQ(term: Term): string {
    switch (term.termType) {
      case BlankNodeTermType:
        return '_:' + term.value
      case DefaultGraphTermType:
        return ''
      case EmptyTermType:
        return '<http://www.w3.org/1999/02/22-rdf-syntax-ns#nil>'
      case LiteralTermType:
        return Literal.toNT(term as Literal)
      case NamedNodeTermType:
        return '<' + term.value + '>'
      default:
        throw new Error(`Can't serialize nonstandard term type (was '${term.termType}')`)
    }
  },
```

timbl:
Fixed it in `outliner-bug` branch but I can't  push the branch as it (like the master branch) fails `npm test`

timbl:
Overrode it  so here is the PR https://github.com/solid/solid-panes/pull/214

PR #214, opened 2020-02-03
removed unused code 

timbl:
partly to  solve https://github.com/solid/solid-panes/issues/213


timbl:
(The code with YAHOO was code to use an old Yahoo drag and drop library. That is not used and could be removed.  That removed on  file too.)

michielbdejong:
That commit I just added fixes one of the two problems ^
Travis now gives one  ✔  and one ✘.
I'll look into the other one.

package.json
```diff
@@ -89,7 +89,7 @@
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged",
-      "pre-push": "npm test"
+      "pre-push-disabled": "npm test"
```
michielbdejong: @timbl can you try to run `npm test` now, maybe run `npm install meeting-pane@2.0.0` in your local checkout? If you still see failing tests, can you paste the log?

michielbdejong (changes requested):
Agree with all the removed unused code and added two commits to fix the Travis build which was broken in two ways, but I think the `pre-push` -> `pre-push-disabled` should not be necessary.

package.json
timbl: Now passes tests.  Thanks!

michielbdejong (approved):

timea-solid:
see PR https://github.com/solid/solid-panes/pull/340 as a switch of this work

PR #215, opened 2020-02-07
Fix Travis build

michielbdejong:


michielbdejong:
The travis - github api key was accidentally removed, we can uncomment this as soon as it's restored.

Vinnl (approved):

PR #230, opened 2020-04-30
Removing unnecessary castings and fixing types

megoth:
Relies on https://github.com/linkeddata/rdflib.js/pull/415 and https://github.com/solid/solid-ui/pull/304


timbl (approved):
Looks good to me.  Nice to see LiveStore creep into the picture.  So good to get rid of all the type castings.

PR #236, opened 2020-07-24
Remove webpack build from main build task

michielbdejong:


michielbdejong:
I think since we're doing a webpack build in mashlib, the one in solid-panes is not really necessary?

megoth (approved):
I think this should be safe, but note that I'm a bit out of date wrt the current build system with mashlib and related repos.

You should also double-check with @timbl whether he uses the bundled version that `build-dist` produced in any of his projects. (But in any case, it should be enough with the developer/unminified version of that bundle anyway, I think.)

angelo-v:
I don't think we should remove it. Usually node_modules are not transpiled again, and I think solid-panes should be a usable on it's own and not only within mashlib, shouldn't it?

angelo-v:
Ah ok, the module build is just done with babel directly without webpack, ok then it should be fine

angelo-v:
Ah ok, the module build is just done with babel directly without webpack, ok then it should be fine

PR #242, opened 2020-08-14
install and register activitystreams-pane and react dependencies

angelo-v:
There is lot of work to be done on the pane itself, so there will be coming some version updates. We could merge it anyway or leave the PR open for a while until the pane is in a state we are happy to release


angelo-v:
About the icon: I just picked one from solid-ui at will... Not sure if this is meant to represent something else or which icons I could use or not. I propable should create a new one?

michielbdejong (approved):
\o/

PR #244, opened 2020-08-17
Activitystreams pane 0.3.0

angelo-v:
this is the latest version, introducing styles and published data


michielbdejong (approved):
No need for code review on chores like this, I would say

PR #245, opened 2020-08-19
Chat with me button on profile

michielbdejong:


michielbdejong:
Depends on https://github.com/solid/chat-pane/pull/45

michielbdejong:
Idea for a follow-up improvement: The current implementation requires the logged in user to grant the visited pod Write+Control access to their own pod, which is not ideal. To resolve this, we should have a list of deployed chat apps that can receive a 'create' action, and see if the user has any of those as a trusted app. If they don't, the button can still offer to send the other user an invite, without creating the chat room (leaving that up to the other person)

michielbdejong:
Screenshot 2020-08-20 at 12 02 37
Screenshot 2020-08-20 at 12 02 53
Screenshot 2020-08-20 at 12 03 13
Screenshot 2020-08-20 at 12 03 53

SharonStrats (approved):
I was able to test this on mashlib-dev. Pretty cool 👍

src/profile/profile.view.ts
```diff
@@ -79,6 +80,31 @@ const thisPane: PaneDefinition = {
...
+          dom,
+          undefined,
+          'Chat with me',
+          async () => {
+            try {
+              const chat: NamedNode = await getChat(subject)
+              chatContainer.innerHTML = ''
+              chatContainer.appendChild(longChatPane.render(chat, context, {}))
+            } catch (e) {
+              window.alert(e.message)
```
timbl: I'd tend to do something like
```
chatContainer.appendChild(UI.widgets.errorMessageBlock(dom, e.message))
```
so  that the error message  is left in the workspace and is not modal.
(Would be good to add a close 'x'  to errorMessageBlock that said)

timbl (changes requested):

src/profile/profile.view.ts
```diff
@@ -79,6 +80,31 @@ const thisPane: PaneDefinition = {
 
       // Todo: only show this if there is vcard info
       heading('Contact')
+
+      const chatContainer = dom.createElement('div')
+      const exists = await getChat(subject, false)
+      if (exists) {
+        chatContainer.appendChild(longChatPane.render(exists, context, {}))
```
angelo-v, line 87: Can we place the chat into some kind of sidebar? Right now the actual profile moves down below the chat. Or shall we tackle this with the general redesign of the profile page?

src/profile/profile.view.ts
michielbdejong: https://solid.github.io/solid-ui/examples/error/

src/profile/profile.view.ts
michielbdejong: Yes, for the moment I think it's fine that the chat appears in the place where you click the button. I guess when you come back to a profile where you click 'chat with me' it should maybe show 'reopen the chat you started earlier' so that the actual profile doesn't move down too much. Follow-up issue: https://github.com/solid/solid-panes/issues/246 - let's merge this now and continue that work there.

PR #253, opened 2020-10-12
Document Solid Focus

michielbdejong:


michielbdejong:
@NoelDeMartin please review?

michielbdejong:
Added a commit that shows the difference in format between @NoelDeMartin's [Solid Focus](https://github.com/noeldemartin/solid-focus) and @timbl's [Solid OS Issue Tracker](https://github.com/solid/issue-pane).

NoelDeMartin:
Yes, that describes how Solid Focus stores tasks at the moment. One improvement I'd note is that lists/tasks relationships are defined by a `http://purl.org/vocab/lifecycle/schema#TaskGroup` and `http://purl.org/vocab/lifecycle/schema#task` link. The relationship between workspaces/lists is implied by `ldp:Container` and `ldp:contains`, so that'll probably need to improve at some point. From a POD's perspective, I guess workspaces don't exist and they are just "folders".

But I'm not sure why that's relevant for this repo, because Solid Focus is not a Solid Pane. The document where you added this says "This document describes how the databrowser stores your data as triples in RDF documents on your pod."

michielbdejong:
True, I should change the title and move it out of solid-panes.
In fact, I even want to document data formats like this at the [PDS Interop](https://pdsinterop.org) level, i.e. even comparing across Solid and other personal data stores.

We can discuss it in the Solid OS weekly call tomorrow.

michielbdejong:
Now at https://pdsinterop.org/conventions

PR #261, opened 2020-12-07
remove temporary workarround for npm 7

angelo-v:
Should work without the workarround now

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
-      - run: npm install -g npm # for issues with executables on npm 7
       - run: npm ci
       - run: npm test
       - run: npm run build


RubenVerborgh:
> Should work without the workarround now

Why? Has npm been updated?

angelo-v:
No, I updated the package lock

RubenVerborgh:
That means all devs have to be on npm 7 though. Is that a good idea?

angelo-v:
Are you sure that is a prerequisite?

RubenVerborgh:
Older npms will just overwrite the lockfile and break things again.

angelo-v:
I just tested it with npm 6: test, lint, build and ci run trough fine

angelo-v:
> Older npms will just overwrite the lockfile and break things again.

That's a general issue with mixing npm versions isn't it?

RubenVerborgh:
> That's a general issue with mixing npm versions isn't it?

Yep, hence the current setup.

angelo-v:
So your concern is, if someone with npm 6 overrides the current lock file the build would break again, if we merge this?

RubenVerborgh:
Yes; given that we are not mandating npm 7 for this project, I assume everyone is running npm 6.

angelo-v:
Ok, I tested this as well:

1. `npm install` using npm 6 --> this changes the lockfiles
2. delete node_modules & then with npm 7:
  - `npm ci`
  - `npm test`
  - `npm run build`

It runs without error, the former problems do not occur

angelo-v:
If you wish I can add a lockfile generated from npm 6 into this PR

timea-solid:
This was already updated.

PR #265, opened 2020-12-11
add tests for outline manager

angelo-v:
This PR adds some tests for outline manager:
  * verify that TDs are rendered correctly for NamedNode and Literal
  * verify that NamedNode cells can be expanded

I needed to disable test coverage for the files under test, because obviously this is only a start and we can not meet the actual test coverage goal for those old files


src/outline/manager.test.ts
```diff
@@ -0,0 +1,99 @@
+/* eslint-env jest */
+import OutlineManager from './manager'
+
+import { lit, NamedNode, sym } from 'rdflib'
+import { findByText, getByText } from '@testing-library/dom'
```
SharonStrats, line 5: I wasn't sure about using testing-library?  I think we may have tried to stay with jsdom.  I can't remember why exactly.  This is something to look at though, because potentially if we can use the testing-library/dom there may be a chance I could use it in solid-ui for draganddrop because that is not supported with jsdom.

SharonStrats (approved):

src/outline/manager.test.ts
angelo-v: jsdom is not a testing library, just a node implementation of dom. @testing-library/dom is imho the perfect fit to test our functionality so I do not see why we should not use it.

Issue #263, opened 2020-12-10
default pane: Add link to each resource


angelo-v:
The default pane ("about" tab) renders a tree of linked resources. The resources can be expanded and collapsed, but there is no direct link to it. This is especially annoying when fetching the expanded resources fails (e.g. due to CORS error).

The linked resource has to be openend in a new browser tab, which can be done by dragging it to the URL bar, but this is not as convenient as just clicking a link and I guess many people are not aware of that feature at all.

I propose to add a link icon like that, next to the resource label, that allows to open the resource in a new tab:

![image](https://user-images.githubusercontent.com/340725/101806484-4f415b00-3b14-11eb-9059-112add27769d.png)

michielbdejong:
@angelo-v so what was the conclusion from the discussion in the meeting just now?

angelo-v:
Let's do it and improve from there on

PR #267, opened 2020-12-18
add link icon for named nodes in default outline

angelo-v:
This PR adds a link icon next to each named node in the default view, as described in https://github.com/solid/solid-panes/issues/263

Preview:


SharonStrats (approved):
Looks and works great!

PR #272, opened 2021-01-07
WIP: PoC building a profile page using ionic-components

angelo-v:
I did an experimental profile page using ionic components and lit-html

This is how it looks:

image

I think making use of ionic would be good to be able to have a clean, responsive design without putting too much efford into an own design system first.

lit-html is another good opportunity to have html templating without going too far away from vanilla js. Existing solid-ui components can be integrated as you can see with the friend list in this example, which is still the same.

### DIFF
--- a/dev/index.html
+++ b/dev/index.html
@@ -1,28 +1,24 @@
 <!DOCTYPE html>
 <html>
-  <head>
-    <meta charset="utf-8" />
+<head>
+    <meta charset="utf-8"/>
     <title>Solid Pane Tester</title>
-    <style>
-      body {
-        margin: 10%;
-      }
-      input {
-        background-color: #eef;
-        padding: 0.5em;
-        border: 0.5em solid white;
-        font-size: 120%;
-      }
-    </style>
-  </head>
-  <body>
-    <h1>Solid Pane Tester</h1>
-    <div id="loginBanner"></div>
-    <p>
-      A handy tool for pane developers. Put your JS or TS file in dev/pane/.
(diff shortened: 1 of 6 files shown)


michielbdejong:
As discussed in today's meeting, this looks great!
We've been talking about making Solid OS more beautiful for years, but most attempts so far keep grinding to a halt.
So let's use something like this for the public view (not logged in) of someone's webid profile. There, we don't have the restriction that we want the pane to be nestable, and we don't need data fields to be editable, so it's much simpler.

Sounds from what you said in the meeting like the main hurdle is working around the problem that Ionic wants to be full-page. Maybe put it inside an iframe?

angelo-v:
I don't think we should introduce something like ionic, if we are not willing to use it as a base component library for SolidOS in general. We need a clean, consistent, responsive UI everywhere, not only on the profile. If not with ionic, then with something else (self-written if it has to be).

> There, we don't have the restriction that we want the pane to be nestable, and we don't need data fields to be editable, so it's much simpler.

Why would nesting or having editable data fields be problematic with ionic?

I am doing another experiment using tailwind now, but I think it is leading in a similar direction: We need to decide what to use as a base for SolidOS design.

michielbdejong:
If you're the person improving the look and feel, and TailWind is what you prefer to use, then unless it undoes or blocks some work from someone else, then I would say "Wer baut hat Recht"! :)

angelo-v:
But this is the main issue: I cannot make such a fundamental decision just for the profile page. This is something that effects SolidOS at whole and how we proceed with UX/UI.

PR #275, opened 2021-01-14
WIP: Tailwind profile poc

angelo-v:
Similar to #272  I did an experimental profile page using tailwind css and lit-html

This is how it looks:

image

It is also looking quite nice and it is responsive. As @NoelDeMartin said in the call tailwind is much more low level and does not bring own components, so profile html is a bit more complex. Therefore it allows greater flexibility, while not needing to write every css rule from scratch.

But, similar to the ionic approach using tailwind would also mean a commitment to that. Tailwind does some CSS normalization and it does not make much sense to only use it for profile page. It also uses global CSS which was something we wanted to avoid in favour of inline styles iirc.

Opinions please :)


NoelDeMartin:
Something I didn't mention in the call is that there is actually a sort of component library for tailwind, and that is [Tailwind UI](https://tailwindui.com/). It is interesting because it also gives the developer full control, given that it consists on copy-pasting html that can be tweaked. They are also working on [Headless UI](https://headlessui.dev/) for javascript components (although that's still WIP).

As much as I personally like Tailwind, I'm not sure if it's the way to go either. But I think it's a very good project to look for inspiration, because I think the flexibility they give is key. You can really build "anything" with tailwind. You can also build anything with Ionic or Bootstrap, but it's a lot more difficult and it feels like you're working against the framework, not with the framework.

I hope to have more to say on the topic once I try to implement some panes :).

megoth:
An article on some of the negative aspects of Tailwind: https://dev.to/jaredcwhite/why-tailwind-isn-t-for-me-5c90

(Let me just say that I've enjoyed using Tailwind (so far), so wouldn't object to use it in SolidOS. Just thought I should share this article to enrich the discussion.)

PR #276, opened 2021-01-18
WIP: PoC building a profile page using lit-html and inline styles

angelo-v:
Following my experiments in https://github.com/solid/solid-panes/pull/272 and https://github.com/solid/solid-panes/pull/275 I created another profile PoC using just lit-html and inline styles (using lit-html's styleMap).

Although it is much more work to go this path, it might be the way to go to allow us most flexibility and independence of third-party libs.

I outlined how we could provide a set of base styles (exported e.g. from solid-ui) that can be combined to create more complex styles / components. This is inspired by the way tailwind provides relatively low-level utility classes.

This approach should also match / be compatible with the ideas you @timbl described in the design issue you shared a while ago, am I correct?

Going forward with this would mean to add (only) lit-html as a third-party dependency. I think this is fine, since it is very light weight and does not effect other panes. We might continue to use it elsewhere, if we see fit, but it is also fine to not do so.

/cc @noeldemartin

### DIFF
--- a/dev/index.html
+++ b/dev/index.html
@@ -1,28 +1,21 @@
 <!DOCTYPE html>
 <html>
-  <head>
-    <meta charset="utf-8" />
+<head>
+    <meta charset="utf-8"/>
     <title>Solid Pane Tester</title>
-    <style>
-      body {
-        margin: 10%;
-      }
-      input {
-        background-color: #eef;
-        padding: 0.5em;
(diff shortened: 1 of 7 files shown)


angelo-v:
This is what it looks, btw. 

image

Even with inline styles I managed to get it at least a bit responsive using grid layout. So the two colums swap into one on small screens

PR #277, opened 2021-01-22
fix XML prefix

angelo-v:
the examples all use `XML:` and not `xsd:`


TallTed:
@angelo-v — I think your "fix" went in the wrong direction, even if it was the fastest, smallest change.

See http://prefix.cc/xsd vs http://prefix.cc/XML

Also https://www.google.com/search?q=xsd vs https://www.google.com/search?q=XML

Yes, technically anyone can use virtually anything for a prefix, but when there's one that's well known and commonly used elsewhere... that's a better choice for things meant to be consumed by anyone and everyone, such as these examples.

angelo-v:
I think XML: is what is generated by rdflib. I did notninvent it neither am I a fan of it. So feel free to change.

Would be cool if we could make rdflib use established prefixes as well

PR #278, opened 2021-01-25
replace profile pane with solid/profile-pane

angelo-v:
see https://github.com/solid/profile-pane

See my WebID as a demo deployment https://angelo.veltens.org/profile/card#me


michielbdejong (approved):
Maybe add a note at the top of src/profile/profile.view saying that it's no longer used?

michielbdejong:
\o/

angelo-v:
> Maybe add a note at the top of src/profile/profile.view saying that it's no longer used?

oh, sorry, just forgot to remove that code

PR #280, opened 2021-02-28
Pass solid logic to pane

angelo-v:
finally pass solid logic to the panes via data browser session

This PR requires https://github.com/solid/pane-registry/pull/11 to be merged first, and pane-registry 2.4.0 to be available

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "solid-panes",
-  "version": "3.4.8",
+  "version": "3.5.0",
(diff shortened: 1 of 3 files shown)


michielbdejong (approved):
Same remark as in https://github.com/solid/pane-registry/pull/11#pullrequestreview-600499482 but we can remove the `.store` and change `UI.store` -> `UI.logic.store` everywhere (this will probably be in hundreds of places) in a follow-up.

PR #284, opened 2021-03-23
render schema:image properties as image

angelo-v:
I also extracted the relevant part from manager.js and added tests for it


michielbdejong (approved):
\o/

PR #291, opened 2021-05-23
Curriculum vitae in public profile form

timbl:
This uses the autocomplete in the latest solid-ui


timbl:
I've added a 'weight' parameter to the Group from field to be able to distinguish groups which are not significant and may  be just be used to help organize the form and those which we want to emphasis to the user with a border around them.

SharonStrats (commented):
It doesn't seem like there are too many changes, and the changes look good to me.  Nothing stands out, however I can't test it.  This branch seems to need something different from contacts-pane. I get two errors in contacts-pane/individual.js one about not resolving /lib/forms and one about lib/vcard.js.  I will have to look at this tomorrow, sorry about that.

bourgeoa (approved):
LGTM tested locally

PR #296, opened 2021-08-07
Fixed linters and tests

jaxoncreed:
Other Test and Linter Fixes:
 - Mashlib (https://github.com/solid/mashlib/pull/121)
 - Solid-Panes (https://github.com/solid/solid-panes/pull/296)
 - Solid-UI (https://github.com/solid/solid-ui/pull/421)
 - Rdflib (https://github.com/linkeddata/rdflib.js/pull/511)


src/dataContentPane.js
```diff
@@ -177,7 +177,7 @@ module.exports = {
             .statementsAsTables(obj.statements, context)
           res.setAttribute('class', 'nestedFormula')
           return res
-        case 'Variable':
+        case 'constiable':
```
bourgeoa: This change is not correct

bourgeoa (changes requested):

src/dataContentPane.js
jaxoncreed: Thanks for the catch

PR #299, opened 2021-08-26
Modification for Data-Kitchen

jeff-zucker:
If the user or app has set a webId for no-auth viewing and we are viewing a no-auth podroot, show the dashboard without checking for login status. For example, now the user's view of their private local pod will have the same dashboard as the logged in user - with pointers to their local profile, settings, and preferences and "MyStuff" links.  This does *not* implement my previous suggestion of letting the app create the home page, it simply treats a local home page the same as a logged in user.


jeff-zucker:
Sorry, needs fixing, will resubmit.

PR #300, opened 2021-08-26
minor modification for Data-Kitchen-like scenarios

jeff-zucker:
Check authn.currentUser because we might be authorized to view  a private dashboard without authn.authSession.info.webId being set.


jaxoncreed (approved):
LGTM

PR #302, opened 2021-09-10
Schedule hours

timbl:
Responding to https://github.com/solid/meeting-pane/issues/28 the problem that the scheduler pane only allows you to pick days, not to pick times.


timbl:
I'm going to merge this mainly as no one else seems to be using this facility and it is is only a form change, I'll risk it not  working and test it with the auth upgrade stuff.

Issue #71, opened 2019-05-06
FolderPane is not refreshed properly after deleting a resource


megoth:

Vinnl:
I think this could theoretically be resolved by adding a change listener (e.g. `kb.updater.addDownstreamChangeListener`) for every object in a folder. node-solid-ws should send a notification through WebSockets that a file has been deleted, but after debugging it a bit with @michielbdejong, it appears that that does not work due to the following bug: https://github.com/solid/node-solid-ws/issues/4

Thus, I think this issue is blocked by that one.

timbl:
Surely the folder pane only has to listen to changes in the folder resource itself, not all the children?

timea-solid:
Deleting a resource is triggered in the internal-pane part of solid-panes, which in turn should affect the display of the immediate, parent folder, displayed in the folder-pane. So only the div of that resource has to be rerendered, which is part of the folder-pane.

timbl:
In this bug, how was the resource deleted?   Using the internals pane, as the folder pane itself does not have a delete function.  Assume something out of the pane itself.   So we have to listen for a change.

timbl:
Fixed by https://github.com/solid/folder-pane/pull/23

PR #304, opened 2021-09-24
Refresh view after deleting folder

timea-solid:
fix attempt for https://github.com/solid/solid-panes/issues/71
This fix goes across folder-pane too.
Together with PR: https://github.com/solid/folder-pane/pull/23
In the code I am making use of dom.closest() function which is compatible with all browsers accept IE which I think is ok: https://developer.mozilla.org/en-US/docs/Web/API/Element/closest


timbl (changes requested):
We should not need this, as the view should be watching the containing folder and automatically get updated when anything is changed in it.  Suggest no changes to this file.

timea-solid:
Was fixed better with: solid/folder-pane#23

PR #305, opened 2021-10-01
fix(manager): resolved menu select going to the correct tab

SharonStrats:
In the tabWidget in solid-ui it is looking for dataset.name to select the tab instead of globalPaneName.  This small change corrects the problem where when you select a menu item it would not go to the correct one you selected but instead went to the first one no matter what.

When I'm on the dashboard and select the menu that appears in the header, I select "Edit My Profile" I should be taken directly to "Edit My Profile" instead of "Your Stuff"


timbl (approved):
Makes sense to me. Have not tried it.

timea-solid:
I just tried it and it works already way better! One thing is still happening: the menu item "show your profile" does not bring me anywhere.

timea-solid:
I also have a weird behaviour in addition. The tab is opened in place
Screenshot 2021-10-06 at 17 25 48
I clicked on 'Your profile' and then on 'Preference' 
Both clicks came consecutively from the menu.
To reproduce:
1. Login
2. click on 'Your profile' in menu
3. click on 'Preference' in menu again and you will see the screenshot above. 
Tested on Chrome & Firefox (on main branch - before auth-upgrade)

timea-solid (approved):
tested locally and the menu works perfectly.

PR #306, opened 2021-10-05
Fix homepage link to files

jeff-zucker:
Currently a visitor not logged in sees a link to "user X's files" at user X's pod root.  The link is currently relative which breaks when the databrowser is not served as a server-front end (e.g. as a stand-alone webapp or in Data Kitchen).  This changes it to a full absolute URL.

### DIFF
--- a/src/dashboard/homepage.ts
+++ b/src/dashboard/homepage.ts
@@ -30,7 +30,7 @@ function createDataSection (name: string): HTMLElement {
 
   const publicDataLink = document.createElement('a')
   publicDataLink.classList.add('list-group-item')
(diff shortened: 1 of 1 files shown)


timea-solid:
I tried to test this locally but I am not sure how to because on localhost I cannot get to the homepage.ts page -> it always links me to the server page instead.

bourgeoa:
@theRealImy You can try it with multiuser locally. The only point is that logout do not really log you out in that case

timea-solid:
Multiuser setup helped. I could successfully test this. It works.

PR #310, opened 2021-11-01
add window.location storage to dashboard

bourgeoa:
Dashboard will list storages from webId and from window.location pod.
see issue https://github.com/solid/solidos/issues/77

### DIFF
--- a/src/outline/manager.js
+++ b/src/outline/manager.js
@@ -415,6 +415,30 @@ export default function (context) {
       .concat(pods)
 
     async function getPods () {
+      async function loadContainerRepresentation (pod) { // namedNode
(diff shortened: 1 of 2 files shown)


src/outline/manager.js
```diff
@@ -415,6 +415,30 @@ export default function (context) {
       .concat(pods)
 
     async function getPods () {
+      async function loadContainerRepresentation (pod) { // namedNode
```
timbl, line 418: - Why load and parse the container rather than fetcher load method?

timea-solid (approved):
Tested locally and it works as described by Alain.

src/outline/manager.js
bourgeoa: Because you did it in first place and I need access to container Representation.
Does fetcher.load() know if the representation in store is the `Accept: text/turtle` ?

bourgeoa:
@timbl I'm merging for the time being. I can update the code to fetcher.load() later.

PR #312, opened 2021-11-09
allow folder-pane to be displayed for container with index.html

bourgeoa:
- add container representation to activate folder-pane
- keep -Your Storage- as name for current webId storage
- clear unused sym

### DIFF
--- a/src/dashboard/dashboardPane.ts
+++ b/src/dashboard/dashboardPane.ts
@@ -1,5 +1,5 @@
 import { authn, icons, store } from 'solid-ui'
-import { Fetcher, NamedNode, sym } from 'rdflib'
+import { Fetcher, NamedNode } from 'rdflib'
 import { generateHomepage } from './homepage'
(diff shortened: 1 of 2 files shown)


jeff-zucker:
I'm sorry, I'm a bit lost.  I do not see any commits on browse.html.  What does this PR have to do with browse.html?

bourgeoa:
@jeff-zucker 
The list of panes displayed depends on the url.
A container with index.html is seen as a resource, not as a container. We need a dual representation. folder-pane is made available when a container representation exist in the store.
The PR loads the container representation in store when if it is missing. The folder-pane can  be listed along side dokieli and human-pane for html.

We have the dual representation for all containers with index.html including root container for which default index.html is served.

jeff-zucker:
@bourgeoa that all makes sense in relation to *index.html* but the title refers to *browse.html* and that is the part I do not understand.  Browse.html is the web app.  Does this commit change browse.html in any way?

bourgeoa:
@jeff-zucker I changed the title.

PR #314, opened 2021-11-14
get dashboard for webApp

### LINKED ISSUE #89: Webapp only display dashboard when webId and search url have same origin. Search uri should be used
webId and search url have same origin. Url storage is used search uri should be used.
image

webId and search url have different origin
image

bourgeoa:
resolves https://github.com/solid/solidos/issues/89


jeff-zucker (approved):
Works great, thanks!

PR #316, opened 2021-11-16
github markdown displayed in human-pane

### LINKED ISSUE #315: Simplify panes and display github markdown flavor in human-pane
The actual markdown-pane do not display the full markdown flavor.
A panes simplification can be achieved displaying markdown directly in human-pane

bourgeoa:


src/humanReadablePane.js
```diff
@@ -86,6 +81,28 @@ module.exports = {
...
     }
 
+    //  @@ When we can, use CSP to turn off scripts within the iframe
+    div.setAttribute('class', 'docView')
+    const iframe = myDocument.createElement('IFRAME')
+    let dataUri
+
+    // render markdown to html
+    const markdownHtml = function () {
+      kb.fetcher.webOperation('GET', subject.uri).then(response => {
```
angelo-v, line 91: **question** why do we need another fetch here? The resource should already have been fetched, hasn't it?

bourgeoa: The original fetch is a HEAD only to find the contentType. We need the content now to feed the `data:uri`.

src/humanReadablePane.js
```diff
@@ -86,6 +81,28 @@ module.exports = {
...
+        const res = marked.parse(response.responseText)
+        dataUri = 'data:text/html;charset=utf-8,' + encodeURIComponent(res)
+        iframe.setAttribute('src', dataUri) // allow-same-origin
```
angelo-v, line 94: **suggestion (non-blocking)** you could just render the markdown html into a div, no need for an iframe in this case

bourgeoa: I decided to keep the same html display for both. It is a lazy solution. The performance is quite good.

angelo-v: I am fine with that

angelo-v (approved):

PR #317, opened 2021-11-30
Markdown div

bourgeoa:
replaces markdown display html in iframe by markdown display html in DIV to allow the opening of external links in a new window.


timea-solid:
I just tested locally with your branch. It seems to be working ok.

timea-solid (approved):

PR #320, opened 2021-12-10
Fix e.indexOf error and update webpack config

### LINKED ISSUE #319: form pane throws error e.indexOf is not a function
The form pane fails render for, e.g. #Object
```
@prefix ui: <http://www.w3.org/ns/ui#> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

<#Object> a <#Class>.

<#Class> ui:annotationForm <#MyForm2> .

<#MyForm2>
    a ui:SingleLineTextField ; 
      ui:label "name" ;
      ui:property foaf:name
```

Error is:
```
Uncaught (in promise) TypeError: e.indexOf is not a function
    at o (uri.js:35)
    at Object.t.document (uri.js:50)
    at pane.js:76
    at fetcher.js:1017
```

This occurs because a NamedNode is passed where a string is expected:
https://github.com/solid/solid-panes/blob/6ca75b8fdeb4f70744b59e574d7afe401554dd8b/src/form/pane.js#L93
https://github.com/linkeddata/rdflib.js/blob/1a90b03cef433f456d518e16440fb0281368c445/src/uri.ts#L33
https://github.com/linkeddata/rdflib.js/blob/1a90b03cef433f456d518e16440fb0281368c445/src/uri.ts#L19

josephguillaume:
Closes #319 where $rdf.Util.uri.document expects a string not a NamedNode.

Also updates webpack, which requires the parameter 'static' instead of 'contentBase'

### DIFF
--- a/src/form/pane.js
+++ b/src/form/pane.js
@@ -90,7 +90,7 @@ module.exports = {
           const heading = dom.createElement('h4')
           box.appendChild(heading)
           if (form.uri) {
-            const formStore = $rdf.Util.uri.document(form)
+            const formStore = $rdf.Util.uri.document(form.uri)
             if (formStore.uri !== form.uri) {
               // The form is a hash-type URI
               const e = box.appendChild(
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -38,7 +38,7 @@ module.exports = [{
     '@trust/webcrypto': 'crypto'
   },
   devServer: {
-    contentBase: path.join(__dirname, 'dist'),
+    static: path.join(__dirname, 'dist'),
     compress: true,
     port: 9000
   },


jeff-zucker:
@josephguillaume - the general practice is to pass NamedNodes, perhaps this should be changed at the other end to accept either string or node.  Where is $rdf.Util.uri.document() ?  I can't find it.

josephguillaume:
Happy to follow your guidance.
As listed in the linked issue, there's up to two functions in rdflib.js that would need to be changed.
They have both been explicitly typed as having a string input.
https://github.com/linkeddata/rdflib.js/blob/1a90b03cef433f456d518e16440fb0281368c445/src/uri.ts#L33
https://github.com/linkeddata/rdflib.js/blob/1a90b03cef433f456d518e16440fb0281368c445/src/uri.ts#L19

josephguillaume:
The webpack config has now also been listed as an issue, so this PR also closes #321

jeff-zucker:
Looking at the source of those two methods, it's clear they need a string, so I withdraw my comment.

jeff-zucker (approved):
LGTM

josephguillaume:
Thanks!

PR #323, opened 2021-12-21
Move code of header and footer out of mashlib

timea-solid:
since solid-panes is the lib that creates the panes, the first SolidOS page is on its own like a pane.
Solid-ui elements/widgets should be used through solid-panes. 

Initial issue: https://github.com/solid/mashlib/issues/107

This module depends on solid-ui PR:  https://github.com/solid/solid-ui/pull/461 (the reason why CI fails)
Remark: first commit solid-ui PR (https://github.com/solid/solid-ui/pull/461)
Remark: this change will not affect anything in production. Only changed on mashlib do!


bourgeoa:
There are constants in the code.
I wonder if these constants could be added in a file parameter defaults (useful for tests). That default can be replaced in mashlib where the main-page is loaded.

src/mainPage/footer.ts
```diff
@@ -0,0 +1,19 @@
+import { IndexedFormula } from 'rdflib'
+import { initFooter } from 'solid-ui'
```
SharonStrats, line 2: question: thinking those GitHub action errors will go away once solid-ui change is in main...

src/mainPage/header.ts
```diff
@@ -0,0 +1,75 @@
...
+*/
+const USER_GUIDE_MENU_ITEM = 'User guide'
+const REPORT_A_PROBLEM_MENU_ITEM = 'Report a problem'
+const SHOW_YOUR_PROFILE_MENU_ITEM = 'Show your profile'
+const LOG_OUT_MENU_ITEM = 'Log out'
+/**
+ * URLS
+ */
+const USER_GUIDE_MENU_URL = 'https://solid.github.io/userguide/'
+const REPOT_A_PROBLE_MENU_URL = 'https://github.com/solid/solidos/issues'
```
SharonStrats: suggestion: I think I would spell it all out REPORT_A_PROBLEM_MENU_URL

src/mainPage/header.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * menu icons
+*/
+const HELP_MENU_ICON = icons.iconBase + 'noun_help.svg'
+const SOLID_ICON_URL = 'https://solidproject.org/assets/img/solid-emblem.svg'
+
+/**
+ * menu elements
+*/
+const USER_GUIDE_MENU_ITEM = 'User guide'
```
SharonStrats, line 11: suggestion: because they are all menu elements, maybe we could have an enum.

SharonStrats (approved):
praise: great job!  Beautiful code!

src/mainPage/footer.ts
timea-solid: yes correct. We have them because this change depends on solid-ui changes to be merged first.

src/mainPage/header.ts
timea-solid: I think enum will not fit with having property files as a next step.

src/mainPage/header.ts
timea-solid: pushed change.

PR #337, opened 2022-02-16
Try move auth

timea-solid:
This is the PR for the move auth from solid-ui to solid-logic.
Something to review: I DO NOT export solid-ui anymore through panes. So integrations like on https://github.com/solid/form-playground/blob/7dd92a5a477c5d23c29d610ba5a957f2834297cc/playground.html#L13 or on solid-ui-components will need to change. Is this acceptable?


jeff-zucker:
I will look when I have time but I find it really bad to drop export of panes. That will break at least a half dozen of my apps and other people's as well.

PR #344, opened 2022-04-11
use authenticated fetch

### LINKED ISSUE #301: Slideshow pane breaks on CSS
CSS does not advertise the media-type of contained resources so when slideshowPane looks for images, it finds none since it is looking in the container for triples listing the item as of type ...iana/image... and CSS does not provide such triples in the container.  I suggest we use the mime-types library to examine the path extension rather than looking for triples.

bourgeoa:


timea-solid (approved):

timea-solid:
Tested it and it works now.

PR #349, opened 2022-04-27
use authenticated fetch in human pane

### LINKED ISSUE #347: Display PDF-Files even if they can only be accessed authenticated
Currently on CSS if a PDF is not publically available SolidOS renders itself (instead of the pdf) in an iframe and shows an error:

image

bourgeoa:


src/humanReadablePane.js
```diff
@@ -104,11 +101,20 @@ const humanReadablePane = {
...
+          const objectURL = URL.createObjectURL(myBlob)
+          frame.setAttribute('src', objectURL) // w640 h480 //
+          frame.setAttribute('class', 'doc')
+          frame.setAttribute('style', `border: 1px solid; padding: 1em; height:120em; width:80em; resize: both; overflow: auto;`)
+        })
     }
-
     // @@ Note below - if we set ANY sandbox, then Chrome and Safari won't display it if it is PDF.
     // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe
     // You can;'t have any sandbox and allow plugins.
```
TallTed:
```suggestion
    // You can't have any sandbox and allow plugins.
```

src/humanReadablePane.js
```diff
@@ -104,14 +101,23 @@ const humanReadablePane = {
     if (ct === 'text/markdown') {
       markdownHtml()
     } else {
-      frame.setAttribute('src', subject.uri) // allow-same-origin
-      frame.setAttribute('class', 'doc')
-      frame.setAttribute('style', 'resize = both; height:120em; width:80em;')
+      // get with authenticated fetch
+      console.log(`humanReadablePane: ${ct}`)
+      kb.fetcher._fetch(subject.uri)
```
angelo-v, line 106: **question** why not `kb.fetcher.webOperation` like with markdown?

bourgeoa: because kb.fetcher.webOperation either return resolve(response.text or resolve(response) so the reponse is already resolved

timea-solid (approved):
Tested locally on NSS and works.

PR #364, opened 2023-01-31
replaced require and module.exports with import and export

timbl:
Also add communities list after friends lst in profile editor.
wa going to do that but found had to progressively spread ESM modulehood, so that became the main change

### DIFF
--- a/src/tableViewPane.js
+++ b/src/tableViewPane.js
@@ -15,9 +15,11 @@
 // ///////////////////////////////////////////////////////////////////
 
 // Table view pane  -- view of a class/
-const UI = require('solid-ui')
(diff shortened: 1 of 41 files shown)


src/tableViewPane.js
```diff
@@ -14,10 +14,8 @@
 
 // ///////////////////////////////////////////////////////////////////
 
-// Table view pane  -- view of a class/
+// Table view pane  -- view of a class asa table of properties of class members
```
TallTed:
```suggestion
// Table view pane  -- view of a class as a table of properties of class members
```

bourgeoa (approved):
LGTM this is nice

PR #368, opened 2023-03-08
sanitize markdown

### LINKED ISSUE #369: XSS in markdown renderer
Split out of SolidOS/solid-panes#353 

The markdown implementation is vulnerable to XSS. When viewing a markdown file, arbitrary scripts from this file can be executed. As an example, go to https://sheep.solidcommunity.net/public/ and click on the xss.md file, which will call print() as a XSS demonstration. The attacker then has the same privileges as mashlib (likely read, write and control).

A POC markdown file that will call `print()` when it is displayed:

```markdown
# Hello

<img src=1234 onerror=print() />
```

A relevant [comment](https://github.com/SolidOS/solid-panes/issues/353#issuecomment-1235842401) from @bourgeoa :

> SoliOS is using https://github.com/markedjs/marked. They recommend to sanitize using https://github.com/cure53/DOMPurify

bourgeoa:


bourgeoa:
There are no tests. Just did : 
- update mashlib with solid-panes@3.5.31-39483f75 and then

- npm run prepublishOnly
- npm run startStaticOS
- created a `test.md` resource with XSS issue using this sanitized example https://github.com/azu/safe-marked#usage

```

<script>alert(1)</script>
<iframe src="https://example.com"></iframe>

This is [XSS](javascript:alert(1))

```

- result is : 
```
<p>This is <a>XSS</a></p>
```
I kept marked and dompurify separate for a better understanding if we ever decide to change one or the other package.

timea-solid (approved):

PR #378, opened 2024-01-13
package.json fix?

chunt007:
On Windows 10 in VSCode, solid-panes will not start and will give you a ./ error, unless you revise the following from

"./timestamp.sh > src/versionInfo.ts  && eslint 'src/versionInfo.ts'

to 

"build-version": "eslint --fix",


README.md
```diff
@@ -77,4 +77,7 @@ Here, just to show how it works, are how some RDF Classes map onto panes. Anythi
 contacts (A VCARD Address Book, Group, Individual, Organization) can be handled by the one contact
 pane. Any other pane which wants to deal with contacts can just use the pane within its own user interface.
 
+If you have trouble seeing syntax highlighting for SPARQL, Turtle, TriG, N-Triples, N-Quads, etc...
+Try installing an extension! For VSCode, there are many, but a great one is RDFox Syntax Highlighting.
```
TallTed, line 81:
```suggestion
If you have trouble seeing syntax highlighting for SPARQL, Turtle, TriG, N-Triples, N-Quads, etc.,
try installing an extension! For VSCode, there are many; one that we've used is 
[RDFox Syntax Highlighting](https://marketplace.visualstudio.com/items?itemName=rdfox.rdfox-rdf). 
```

TallTed (changes requested):

package.json
```diff
@@ -10,7 +10,7 @@
     "build-lib": "npm run build-form && babel src -d lib --source-maps --extensions '.ts,.js'",
     "build-dev": "webpack --progress --mode=development",
     "build-types": "tsc --emitDeclarationOnly",
-    "build-version": "./timestamp.sh > src/versionInfo.ts  && eslint 'src/versionInfo.ts' --fix",
```
angelo-v, line 13: This replaces whatever the build-version script did before by calling timestamp.sh with just a call to eslint. Does not seem right to me.

angelo-v (changes requested):

package.json
chunt007: Not sure as to why VSCode throws off that error, but in the mean time, perhaps we could use it as a cautionary suggestion to Windows 10 users? If it doesn't seem right, then maybe it is not. Should the suggestion instead go the readme?

angelo-v: Currently the dev setup seems not to be Windows compatible. At least the timestamp.sh would need to change if we want to make the build run under windows. There might be more changes needed. Under Windows you could try using WSL.

PR #389, opened 2024-10-28
Add a tab to the dashbaosd with the profile itself (not the editor)

timbl:
When editing your proofile, its good to be able to see the edited pofile. Plus. generally useful.  And the profile mmenu ""go to profile" only works under some conditions.


timbl:
1. A small but useful change to the menu seems to work when  tested

PR #400, opened 2025-06-14
Pane login issues after move to pivot.

chunt007:
https://github.com/SolidOS/solid-panes/blob/main/src/dashboard/dashboardPane.ts and editProfile pane don't seem to act correctly.

Developer console spits this out

Capture
 
 link.href = '/common/css/bootstrap.min.css' seems to be a clue.

temporary remedy was 'http://localhost:8080/src/dashboard' incase you could bypass this error with a local .css bootstrap work around.  for local static http server command is http-server ./. 
This was to bypass internet browser security's attempt to flag a localized file.

Capture2

Capture3

Even with a work around, its still blank.

authSession.onLogin(() => {
      // console.log('On Login')
      runBuildPage()
    })
    authSession.onSessionRestore(() => {
      // console.log('On Session Restore')
      runBuildPage()
    })
    // console.log('Initial Load')
    runBuildPage()

    return container
  }

Within this code, the login window will just keep "rebuilding" until you logout or until this code is removed.

**<!DOCTYPE html>** declaration is also flagged.


chunt007:
Panes rendering the CSS for both Profile Pane and Solid Panes repo without CSS minify issue. User should check for fresh installs of node.js, Chocolatey, and webpack dev. npm audit fix is not guaranteed to work. While highly doubtful, Windows 10 may have underlying package issues with these repositories, but Windows 11 works perfectly fine. No Linux Subsystem for Windows is being used.

PR #563, opened 2026-03-10
fix version

SharonStrats:
There was a problem with version when i tried to run locally and this fixed it.


SharonStrats:
Actually I don't think we need this. I should've just ran build version. sorry.

PR #610, opened 2026-04-18
Milestone3m navbar

bourgeoa:
manager.js — visually hides the tab strip <nav> after tabWidget() is created


timea-solid:
i implemeted this in my work already and imporved the overall code, I am not sure this is needed

PR #611, opened 2026-04-19
Milestone3m navbar - compare against milestone3m

timea-solid:


timea-solid:
The milestone3m branch take a different approach to the navbar - we are not using the tabs pane anymore because we have a left side menu. The challenge there, was to make sure mashlib has the same behaviour when calling GotoSubject for a container or a profile or whatever else. 
The left side menu, when I click on profile should always open the profile regardless if the URI is on a container. The same should happen for storage and for Friends too.

PR #623, opened 2026-04-23
Refactor content rendering to use div containers

bourgeoa:


src/humanReadablePane.js
```diff
@@ -166,6 +166,7 @@ const humanReadablePane = {
 
     // Fallback: detect markdown by file extension if content-type is not text/markdown
     const isMarkdown = ct === 'text/markdown' || isMarkdownFile(subject.uri)
+    const isPlainText = ct === 'text/plain'
```
timea-solid, line 169: Did you take out the markdown files on purpose?

timea-solid (approved):

src/humanReadablePane.js
bourgeoa: markdown has is own rendering function https://github.com/SolidOS/solid-panes/blob/63cacf450e3b6acdcb388590d43bbebfc5a6ee67/src/humanReadablePane.js#L217

PR #719, opened 2026-07-03
New layout according to new design

### LINKED ISSUE #682: Implemnet re-redesign of logged in user menu items
We have a left side menu, when user logged in. We migth need to move this under the header top right.
Also the nav bar menu might move place.

timea-solid:
This PR removes a couple of elements and rearranges the layout: menu moved under logged-in user top right. We do not have a footer anymore.


README.md
```diff
@@ -91,34 +91,4 @@ We have added comments in the code to make it explicit which parts are 100% writ
 
 * Raptor mini: Update the code to use the new solid-ui-header component. Keep in mind the log in and sign up are wired in specific ways.
 
-* Auto: change the menu to fill up the menu items like in the code: async function getMenuItems (outliner: any) {
```
NoelDeMartin, line 94: As I mentioned in another PR, I'm not sure we should remove this. It's still part of the "prompt history". The codebase may have changed, but what has already happpened, can't be changed.

timea-solid: these prompts directly created /were used in the code I deleted. Thus, I believe it's ok to delete the prompts associated with those files too.

NoelDeMartin (approved):

PR #725, opened 2026-07-28
new navbar component

### LINKED ISSUE #722: Redesign nav bar
Implement new design.

timea-solid:
This creates a new Navbar web component.
I removed the top icon tray nav but kept it for the resources in the file explorer for now.


src/components/navbar/Navbar.ts
```diff
@@ -0,0 +1,44 @@
...
+  static styles = styles
+
+  @property({ type: Array })
+  accessor navbarItems: NavbarMenuItem[] = []
+
+  render () {
+    return html`
+    <nav class="navbar">
+       ${this.navbarItems.map(menuItem => html`
+        <button
```
SharonStrats: I have a feeling you are planning to use them later, but just in case you have not used
solid-ui-button or the menu etc..

Also, just a note that I'm going to need to add configuration to the menu-item so that the size of the icon and font can be changed in order to use it for the 3 dot menu. Perhaps you'll need this too.

src/mainPage/navbar.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const podStorages = await getPodStorages(subject.uri)
+  if (podStorages.length === 0) return baseItems
+
+  return [
+    ...baseItems,
+    createNavItem('Storage', () => {
+      const folderPanes = podStorages.map((pod, index) => createFolderPaneItem(outliner?.context?.session?.paneRegistry?.byName('folder'), pod, index))
+      if (folderPanes.length === 0) {
+        console.warn('Folder pane is not registered')
```
SharonStrats: maybe you should also add this message for profile and social.

src/outline/manager.js
```diff
@@ -566,6 +468,24 @@ export default function (context) {
     return getOrCreateContainer('OutlineView', 'Resource browser')
   }
 
+  function getNavbarElement () {
+    return document.querySelector('solid-panes-navbar')
+  }
+
+  function hideSolidPanesNavbar () {
```
SharonStrats: we could use classes instead of inline styles to toggle. Maybe you were going to add that later.

src/outline/manager.js
```diff
@@ -610,26 +530,15 @@ export default function (context) {
...
-  }
-
   async function getRelevantPanes (subject, context) {
     // make sure container representation is loaded (when server returns index.html)
     if (subject.uri.endsWith('/')) { await loadContainerRepresentation(subject) }
     const panes = context.session.paneRegistry
     const relevantPanes = panes.list.filter(
       pane => pane.label(subject, context) && !pane.global
     )
-    if (relevantPanes.length === 0) {
```
SharonStrats, line 629: if there are no relevant panes now what will get shown instead? The folder-pane may need this, I can't be sure yet though just a note to check.

src/outline/manager.js
```diff
@@ -652,7 +561,17 @@ export default function (context) {
   }
 
   async function expandedHeaderTR (subject, requiredPane, options) {
+    async function determineFirstPane (options = {}) {
```
SharonStrats, line 387: I see perhaps it's here, why we don't need the relevant pane ... return 'internal'?

src/outline/manager.js
```diff
@@ -652,7 +561,17 @@ export default function (context) {
...
+      const relevantPanes = options.hideList
+        ? []
+        : await getRelevantPanes(subject, context)
+      tr.firstPane = requiredPane || getPane(relevantPanes, subject)
+    }
+
     async function renderPaneIconTray (td, options = {}) {
+      // Icon tray removed: this UI is no longer needed.
```
SharonStrats, line 395: I'm thinking that this will basically be in that 3 dots menu (at least source pane, maybe others). so anything that was shown here can be accessed still.

src/outline/manager.js
```diff
@@ -847,21 +761,29 @@ export default function (context) {
...
+      )
     }
 
-    header.appendChild(
-      await renderPaneIconTray(td, {
-        hideList: showHeader
-      })
-    )
+    // Pane icon tray removed: preserve first-pane selection without rendering the tray.
+    await determineFirstPane({ hideList: showHeader })
```
SharonStrats, line 596: I'm just making a note here, because I need to recheck this. 

I want to check is this only the tray... because I believe the header also has the file name for instance and I'm going to need to move the tray, so we should technically still have the header but instead of the icon tray they will just be menu items in the 3 dots. 

Basically I think it should still be shown, definitely for a pod "owner" (logged in, looking at your own profile).  I'm not sure about the other 2 viewerModes I need to recheck the design.

i'm kind of thinking that we need the viewerMode that is in profile-pane to be at a higher level here in solid-panes.

src/outline/manager.js
```diff
@@ -1853,15 +1777,15 @@ export default function (context) {
       // Shift forces a refocus - bring this to the top
       outlineRefocus(p, subject, pane)
     } else {
-      if (e.altKey) {
+      /* if (e.altKey) {
```
SharonStrats: If internal is no longer our default, do we have/need a default?

src/outline/manager.js
```diff
@@ -2352,15 +2276,20 @@ export default function (context) {
...
   @param table   -- option  -- default is an HTML table element in which to put the outline.
 */
   this.GotoSubject = function (subject, expand, pane, solo, referrer, table) {
-    table = table || getOutlineContainer('OutlineView') // if does not exist create a compatible host in the current shell
+    const outlineContainer = getOutlineContainer()
+    if (!table || table === outlineContainer) {
+      showSolidPanesNavbar()
+    }
+
+    table = table || outlineContainer // if does not exist create a compatible host in the current shell
```
SharonStrats, line 2036: you probably had to do it for some reason, but funny to move this below the check. Seems like it shoudl stay above it.

src/registerPanes.js
```diff
@@ -113,7 +113,7 @@ export function registerPanes (register) {
   register(tableViewPane)
 
   // Fallback totally generic:
-  register(defaultPane)
+  // register(defaultPane)
```
SharonStrats: this is really interesting. it's actually a Developer only pane, it's the one that says "about".

why would it say for instance fallback totally generic when it could only be shown for someone who has Developer as their user preference.

I think we should discuss what it does today and if we still want it for people who have Developer selected.
Image

src/registerPanes.js
```diff
@@ -139,7 +139,7 @@ export function registerPanes (register) {
   register(sharingPane)
 
   // The internals pane is always (almost?) the last as it is the least user-friendly
-  register(internalPane)
+  // register(internalPane) // under the hood has been removed with new design
```
SharonStrats: Same with this I think we should discuss where the user will see the data now that can show in this pane so we can make sure we don't lose anything.

- the delete we have covered
- the uri we have covered
**- I'm not sure about the .meta though** - I have not looked yet

Image

SharonStrats: Image

package.json
```diff
@@ -68,18 +68,12 @@
...
-    "profile-pane": "3.2.3-3",
+    "profile-pane": "3.2.3-4",
     "rdflib": "2.4.0",
     "solid-logic": "4.0.8-1",
     "solid-namespace": "^0.5.4",
     "solid-ui": "3.1.3-13",
-    "source-pane": "3.1.1-4"
-  },
-  "overrides": {
-    "rdflib": "$rdflib",
```
SharonStrats, line 79: Because rdflib is not in my solidos setup I didn't have to remove this one. Thought I'd just mention.

src/outline/manager.js
SharonStrats: I brought down your code and see that the file name is still there and a few icons like the pane it is, source, and sharing. this is good. so I can just move them to the 3 dots in a different PR that should work, the header is still there.  Interestingly, and I haven't gone through the manager code yet in detail but this change you made did not effect the containers.

Also,  (I am just making notes for consideration). In the new design the containers are instead shown in a square div with the 3 dots at the corner. 

Image

src/components/navbar/Navbar.ts
timea-solid: I tried to use what we have and it did not work. Nabvar is just too different.

src/outline/manager.js
timea-solid: relevant panes exists but is not in use in the menus anymore. the menus are fixed. 
The relevant panes comes into play in the 3 dot menu

src/outline/manager.js
timea-solid: The menu items are not fixed.

src/outline/manager.js
timea-solid: Yes yes.
Manager created the header and icon tray for each resource you open. 
It does this [here](https://github.com/SolidOS/solid-panes/blob/696e63e76208eeff40a35f51ac2fe3e5393bcf69/src/outline/manager.js#L844). I moved the render of the iconTray only when the header exists so that we can still edit files. This is the code you will need to refactor and use in your menu items later.

src/outline/manager.js
timea-solid: Initiall I was thinking there should be no default but I need to think it though better.

src/registerPanes.js
timea-solid: Yes, my thoughts too. This is a pane we should theoretically get rid of I believe.

src/registerPanes.js
```diff
@@ -113,7 +113,7 @@ export function registerPanes (register) {
   register(tableViewPane)
 
   // Fallback totally generic:
-  register(defaultPane)
+  // register(defaultPane)
```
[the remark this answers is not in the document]
timea-solid: I put it back

src/registerPanes.js
```diff
@@ -139,7 +139,7 @@ export function registerPanes (register) {
   register(sharingPane)
 
   // The internals pane is always (almost?) the last as it is the least user-friendly
-  register(internalPane)
+  // register(internalPane) // under the hood has been removed with new design
```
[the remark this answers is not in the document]
timea-solid: I put it back

src/components/navbar/Navbar.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+@customElement('solid-panes-navbar')
+export default class Navbar extends WebComponent {
+  static styles = styles
+
+  @property({ type: Array })
+  accessor navbarItems: NavbarMenuItem[] = []
+
+  private selectItem (item: NavbarMenuItem) {
```
SharonStrats, line 27: it might be better to keep selected parent driven instead of doing it here, just a thought.

SharonStrats (approved):
Looks and works really well, I just left the one comment about the Navbar component. I was able to run it locally as well without any problems.

PR #729, opened 2026-08-01
add and fill 3 dots menu

bourgeoa:
Add 3 dots menu and fill with all panes except the first one (human rendering) and sharing-pane.

There is a bug in solid-ui
> It's a bug in solid-ui-menu: syncItems filters children without slot, then assigns them slot — so the second time it runs, all children have a slot and get filtered out. Let me work around it by using <wa-dropdown> directly:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,7 @@
         "rdflib": "2.4.0",
         "solid-logic": "4.0.8-2",
         "solid-namespace": "^0.5.4",
-        "solid-ui": "3.1.3-13",
+        "solid-ui": "^3.1.3-14",
         "source-pane": "3.1.1-5"
       },
       "devDependencies": {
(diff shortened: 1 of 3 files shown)


SharonStrats (commented):
Thank you @bourgeoa I'm so sorry I don't think I'll be able to use it. I had already did it in a very different way. If I bring it into my branch it will cause me problems I think. I'll have a closer look later and see if I can do anything about this to use part of your work at least.

to give you some context. In source pane I had built the header and I have moved that into solid-panes. the summary (info about file like name) is in one web component and the controls are in another and then there is a header wrapper web component.

I had the same issue as you about the solid-ui-menu. I ended up solving it by changing the pane rerender to only remove the "pane" section not the header. When it was removing the table the menu items got lost and then weren't recreated. That seems to be working now.

bourgeoa:
Thanks @SharonStrats 
I did not find your PR branch work on GitHub.com that's why I did it.

We need an implementation logic for 3-dots menu to stop our weeks discussion about which panes are in and out and no way to see it structurally implemented and pushed to sharing

The panes order is significant.

bourgeoa:
solid-ui-menu issue bug https://github.com/SolidOS/solid-ui/issues/849

PR #737, opened 2026-08-14
Fix bugs and improve styling/mobile for release

### LINKED ISSUE #736: Bugs/Changes Needed for a release
- Remove Delete from menu items as it doesn't work properly yet. Users can still delete from internal pane.
- Would be good to modify the container header to one line and also only 3 dots so it's simplier and less messy on the screen.
- Add mobile styling for the header and some other basic changes so that it looks better on mobile.

Note: the profile pane resize issue where the subject that was showing with data was changing when resizing to mobile was in solid-panes.

SharonStrats:
Implement mobile styling and export file explorer provider and header from index.ts.
Note that most are temporary changes for releasing as the structure will change, so it's just an attempt to match the design as close as possible but taking into consideration the current layout.


src/components/file-explorer-header/FileExplorerHeaderSummary.styles.css
```diff
@@ -12,8 +12,25 @@
 
   h1 {
       color: var(--solid-ui-color-gray-700, #364153);
-      font-size: var(--solid-ui-font-size-2xl, 1.5rem);
+      font-size: var(--solid-ui-font-size-xl, 1.25rem); /* while in expand will go to 2xl when new design is complete */
```
TallTed:
```suggestion
      font-size: var(--solid-ui-font-size-xl, 1.25rem); /* while in expand; will go to 2xl when new design is complete */
```

SharonStrats:
Since we are not releasing we don't need to do this now, it can be done as we go.

src/components/file-explorer-header/FileExplorerHeaderControls.ts
```diff
@@ -8,20 +8,50 @@ import '~icons/lucide/pencil'
...
 
   @property({ attribute: false })
-  accessor menuItems: Array<{ label: string, action: (event: Event) => void }> = []
+  accessor menuItems: Array<{ label: string, action: (event: Event) => void, icon?: HTMLElement }> = []
 
   @property({ type: Boolean })
   accessor canEdit: boolean = false
 
+  @state()
+  accessor isMobile = typeof window !== 'undefined' && typeof window.matchMedia === 'function'
```
timea-solid, line 33: this watchMedia is interesting but I believe this is where maybe we should work with the pane-registry environment. For this release, this is fine for now.

src/components/file-explorer-header/FileExplorerHeaderSummary.ts
```diff
@@ -8,8 +8,9 @@ import { PaneIcon } from './types'
 import '~icons/lucide/globe'
 import '~icons/lucide/lock-keyhole'
 import '~icons/lucide/arrow-left'
+import '~icons/lucide/folder'
 import styles from './FileExplorerHeaderSummary.styles.css'
-import { type FileExplorerResourceMetadata } from './helper'
+import { getContainerItemCount, isContainerSubject, type FileExplorerResourceMetadata } from './helper'
```
timea-solid: in solid-panes we have, brand new, a podUtils unders utils... O believe some functions and logic is now duplicate and yes, it should be in solid-logic actually. We can consolidate later ioni worst case.

src/components/file-explorer-header/helper.ts
```diff
@@ -12,6 +12,25 @@ export type FileExplorerResourceMetadata = {
   modified: string | undefined
 }
 
+export function isContainerSubject (store: LiveStore | undefined, subjectUri: string | undefined): boolean {
```
timea-solid: we could at least move it to podUtils for now

src/components/resource-actions-menu/ResourceActionsMenu.ts
```diff
@@ -1,103 +1,82 @@
...
-  /* private gotoParentFolder(resourceUri: string) {
-    const sourceContext = this.sourceContext
-    const outliner = sourceContext?.context?.getOutliner?.(sourceContext.context.dom)
-
-    if (!outliner) return
-
-    const parentFolderUri = this.getParentFolderUri(resourceUri)
-    ;(outliner as any).GotoSubject(sourceContext.context.session.store.sym(parentFolderUri), true, undefined, true, undefined)
-  } */
+  accessor handleSharingClick: (() => void) | undefined = undefined
```
timea-solid: lets rename sharing to access

timea-solid (changes requested):

src/components/file-explorer-header/FileExplorerHeaderControls.ts
SharonStrats: Yeah I agree, and perhaps move that profile-pane logic. I'll create a ticket.

src/components/file-explorer-header/FileExplorerHeaderSummary.ts
SharonStrats: Yes, I'll leave this for now, but my next PR for the 3 dots menu already uses the resource service from solid-logic. I was just trying to keep this simple for the release and also at the time I wasn't sure you agreed with the resource service.

timea-solid (approved):

PR #741, opened 2026-08-18
Resize fix

SharonStrats:
Fix the issue where on mobile the data gets lost.


timea-solid (approved):
I looked at it in attention. I believe this is a goos imporvement.

PR #749, opened 2026-08-23
Prototype - potential code design for the new design

SharonStrats:
Just playing around with the implementation of the new design to see what we should do and so we can discuss and find missing features.

I have also made lots of comments for now to make it easier to reason about the code and whether we should keep or change it. If we use the work from this branch in the end I will remove all the comments and clean it up.

This is not final and may even be thrown away.

This goes along with the same branch in the folder-pane.

Some questions are how far do we want to go to simplify the OutlineManager. How much do we need to keep in manager and how much can we bring into the folder-pane.

I haven't gotten there yet, but I'm wondering also if we want to bring in more web components ... could gotoSubject functionality be brought into one for instance, this is the main function used by the folder-pane, my gut is it needs to stay in solid-panes but I'm not completely sure yet.

Note: The commits are done in a way to show what steps I've taken so far.

**Additional notes**
Doing away with the table structure, to completely get rid of the table structure there are some other panes that we need to modify. There are 2 functions that are used in other panes appendPropertyTRs and UserInput.
1) defaultPane 
2) internalPane
3) transaction/pane.js
4) ui/pane.js (this is for editing forms i think)
5) classInstancePane
6) argument/argumentPane.js
7) socialPane.ts


timea-solid:
Some of the mentioned panes we will be deleting. TO specify:
* internalPane = under the hood - we will probably delete
* defautPane = about - we will probably delete
These 2 panes will be folded in the 3 dot menu. For now we keep them and delete them wehn the 3 dot menu is in place ans stable.

SharonStrats:
@NoelDeMartin so here, my main question is around the structure and of course anything else you see that stands out. I will create a document in the next day or so to show the new structure so you can see it all at once.

src/mainPage/header.ts
```diff
@@ -91,7 +92,7 @@ export async function createHeader (outliner: OutlineManager) {
   ]
 
   render(
-    html`<solid-panes-header .menuItems=${menuItems}></solid-panes-header>`,
+    html`<solid-ui-provider><solid-panes-header .menuItems=${menuItems}></solid-panes-header></solid-ui-provider>`,
```
NoelDeMartin: This shouldn't be necessary, didn't we already add the `<solid-ui-provider>` wrapper everywhere it was necessary? If not, it should be fixed wherever we find the problem, but the header shouldn't be injecting a `<solid-ui-provider>`.

NoelDeMartin (commented):
I just finished a review of the PR in the solid-ui repo, and I noticed that many of the components I've reviewed as "new code" were actually removed from here 😅. I hope they are still useful though.

Looking at this PR, I'm not sure I can give much feedback because I don't understand a lot about the outline manager and I'd need to spend some time looking into it... Maybe ask me some specific parts of the code where you're interested in my feedback.

Doing a quick look on the changes, I noticed we are adding a bunch of code using imperative APIs (`document.createElement()`, etc.), so I'd try to avoid it, but I'm not sure if that's new code or we've moved it from elsewhere.

Also, regarding the table/div changes in general, and a11y, it's very difficult to review those things just looking at the code. Ideally, I would open a running SolidOS instance somewhere, open a page with the file explorer, and look at the HTML and a11y there. Many times, a11y issues happen because of the way some components are nested or used, but they can look ok on their own.

src/mainPage/header.ts
SharonStrats: you are correct that the dom manipulation you see is old code. I've only used it now very infrequently when it's necessary mixing legacy with new.

